PR #23, opened 2025-07-28
Show vision capabilities in model overview

schicho:
# Show Model Capabilities in Web UI

This is something I was interested in when selecting models to use. This would have helped me to know which models I can use for vision, without trying every model separately.

It adds a new column to the models overview in the aqueduct frontend.

## Implementation

This is a very bare bones PR, as I was mostly looking forward to feedback if something like this was appreciated or even in scope of this project.

This is also not well done from a architecture point of view. I have simply imported the `litellm utils` directly into the view, which is not done anywhere else on the frontend. Also this should maybe extend to other model capabilities. I was unsure where to put the code for this. What is your opinion on this?

Furthermore, I could also not test this as I was running into setup issues, but this is beyond the scope of this PR description.

Please let me know if you are interested in this feature and maybe you can provide hints on how you would design this. In case this is something you would consider, I will get back to this. Alternatively, you can also take over this PR and take this as a user story and implement it way quicker yourself :)


meffmadd:
Thanks for the PR! This is definitely appreciated and within scope!

The problem with the current implementation of this is that `litellm.utils.supports_vision` basically uses a lookup table of popular providers/models (e.g. GPT-4.1). There is no actual logic to determine the vision capabilities. So we have to "hard-code" this somewhere as well.

My first idea would be to use the `model_info` field. See: https://docs.litellm.ai/docs/proxy/model_management 
We could then just put a `supports_vision` field in there and show this in the UI. Let me know what you think!

If you have questions about the setup please send me an e-mail. I am happy to help :)

schicho:
> The problem with the current implementation of this is that `litellm.utils.supports_vision` basically uses a lookup table of popular providers/models (e.g. GPT-4.1). There is no actual logic to determine the vision capabilities. So we have to "hard-code" this somewhere as well.

Yeah I figured that out when I was looking for a way to implement that functionality. It seems they have a quite up-to-date list, however. I agree with you that it's not a perfect solution.

> My first idea would be to use the `model_info` field. See: https://docs.litellm.ai/docs/proxy/model_management We could then just put a `supports_vision` field in there and show this in the UI. Let me know what you think!

Cool, I was looking for something like this, but didn't find any examples thereof. I thought the yaml didn't have further fields. I'll take a look at it. That seems to be a good solution.

meffmadd:
Didn't know that could also be set as a field in `model_info`! Pretty-printing the config might be a bit confusing but I think this only interests "power-users" anyways so might be an option in the future.

Anyways, LGTM!

PR #46, opened 2025-09-23
Add terms of service agreement

meffmadd:


natkam:
Merged without a code review? Tut-tut, I thought we weren't doing this. :grin:

PR #48, opened 2025-09-30
36: Image generation endpoints

natkam:
TODO:
- update openapi.yaml
- do we want to support streaming images? (see [the `stream` param](https://platform.openai.com/docs/api-reference/images/create#images-create-stream), [image streaming guide](https://platform.openai.com/docs/guides/image-generation#streaming))

### DIFF
--- a/aqueduct/gateway/tests/test_image_generation.py
+++ b/aqueduct/gateway/tests/test_image_generation.py
@@ -1,0 +1,194 @@
+import json
+
+from gateway.tests.utils import _build_chat_headers
+from gateway.tests.utils.base import GatewayIntegrationTestCase
+from management.models import Request, Usage
+
+
+class ImageGenerationEndpointTest(GatewayIntegrationTestCase):
+    """Test the image generation endpoint."""
+
+    model = "dall-e-2"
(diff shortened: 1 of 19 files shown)


meffmadd:
Link to the current OpenAI OpenAPI spec: https://github.com/openai/openai-openapi/tree/manual_spec

aqueduct/gateway/tests/test_image_generation.py
meffmadd, line 1: Very nice! At some point we should adapt all tests to use `url` or set the headers in the `setUpClass` etc.

docs/api/image_generation.md
meffmadd, line 1: Did you test the examples?

natkam:
Note: not yet finished

aqueduct/gateway/views/image_generation.py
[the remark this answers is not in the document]
natkam: For future reference:  
After discussing all the possible use cases and edge cases (e.g. multimode models), we decided to drop the validation of mode. It can still be used in the router config as a "comment" - e.g. to document which model supports which mode - but we will not require it.

PR #56, opened 2025-10-10
Batch processing refactor

### LINKED ISSUE #55: Responses are dropped from output/error files in the Batch API
**Describe the bug**
Requests are not written to output files.

**To Reproduce**
Unknown

**Expected behavior**
Once a request is issued, it should always be written to an output file. The batch loop should not be stopped until the request resolves.

Look at batch: `batch-f3ba7bc1e73c472481f708cdd1da37e1`

meffmadd:
Could not reproduce #55 but refactored and added tests


natkam (commented):
I feel like I need more background about what's happening here! Can we talk about it on Monday?

natkam (commented):
A few nit-picky comments because I couldn't find anything really serious that could have caused the bug. Overall it seems to me that the code is more readable after the refactoring!

natkam (approved):

PR #63, opened 2025-10-23
10 mcp server gateway

meffmadd:


aqueduct/gateway/views/decorators.py
```diff
@@ -29,6 +32,92 @@
...
+    - Content-Type header for POST requests
+
+    Returns appropriate status codes:
+    - 421: Invalid Host header
+    - 403: Invalid Origin header
+    - 400: Invalid Content-Type header
+    """
+
+    @wraps(view_func)
+    async def wrapper(request: ASGIRequest, *args, **kwargs):
```
natkam: I would consider splitting this function into smaller ones, each one validating one thing, and with a docstring: one with the host-matching logic, one with the headers validation, etc.

aqueduct/gateway/views/decorators.py
```diff
@@ -465,3 +610,57 @@ async def wrapper(request: ASGIRequest, *args, **kwargs):
...
+            body = request.body
+            logger.error(f"Request body in parse_jsonrpc_message: {body!r}")
+            if body is None:
+                logger.error("Request body is None")
+                return JsonResponse({"error": "Missing request body"}, status=400)
+            data = json.loads(body)
+            json_rpc_message = None
+
+            try:
+                from mcp.types import JSONRPCMessage
```
natkam: Unless there is a particular reason to import it here, I would suggest moving it to the top of the module, where all other imports are (that's also what [PEP8](https://pep8.org/#imports) recommends)

aqueduct/gateway/views/mcp.py
```diff
@@ -0,0 +1,801 @@
...
+                response = JsonResponse({"status": "accepted"}, status=202)
+                if session_id:
+                    response.headers["mcp-session-id"] = session_id
+                return response
+        finally:
+            if operation_registered:
+                session.register_operation_done()
+    except (KeyError, RuntimeError) as e:
+        log.error(f"MCP server '{name}' not found: {str(e)}")
+        return JsonResponse({"error": f"MCP server '{name}' not found: {str(e)}"}, status=404)
```
natkam: The message suggests that you're expecting some very specific error that is not likely to happen in, say, the part of code responsible for handling the situation when `json_rpc_message.root` is a `JSONRPCRequest`. Can you please try to limit the `try ... except` blocks to only contain the exact lines when you expect the given exception to happen? Especially with multiple nested try-excepts, this code is very hard do read. Also, if a `KeyError` happens anywhere else, it's going to be captured here and logged with a message that may be more misleading than helpful (especially given the fact that `str(e)` will just be the key that was not found).

aqueduct/gateway/views/mcp.py
```diff
@@ -0,0 +1,801 @@
...
+    log.info(f"MCP POST {name} - Session: {session_id}, Method: {message_type}")
+
+    try:
+        if is_initialize and not session_id:
+            log.info(f"Initializing session for MCP server '{name}'")
+            mcp_config = get_mcp_config()
+            server_config = mcp_config[name]
+            url = server_config["url"]
+            session_id = await session_manager.create_session(url)
+            log.info(f"Session {session_id} initialized")
```
natkam: Wouldn't this log entry make more sense after we make sure that `session_id` is not `None`?

natkam (commented):
I didn't review everything since you said you were still working on this MR, but I've already left some suggestions for the things that caught my eye.

aqueduct/gateway/tests/utils/base.py
```diff
@@ -126,6 +124,8 @@ def setUpClass(cls):
     def create_new_user() -> tuple[str, int]:
         # Create a new user and a new token for that user
         new_user = User.objects.create_user(username="OtherUser", email="other@example.com")
+        from django.contrib.auth.models import Group
```
natkam: Do we get circular imports with this line? If not, I'd rather put it at the top of the module.

aqueduct/aqueduct/settings.py
```diff
@@ -249,26 +276,26 @@ def batch_processing_concurrency():
...
-        'NAME': os.getenv('POSTGRES_DB', 'aqueduct'),
-        'USER': os.getenv('POSTGRES_USER', 'aqueduct'),
-        'PASSWORD': os.getenv('POSTGRES_PASSWORD', 'aqueduct'),
-        'HOST': os.getenv('POSTGRES_HOST', 'localhost'),
-        'PORT': os.getenv('POSTGRES_PORT', '5432'),
-    })
+if DATABASE_ENGINE == "django.db.backends.sqlite3":
+    DATABASES["default"]["NAME"] = str(BASE_DIR / "db.sqlite3")
+    if TESTING:
+        DATABASES["default"]["TEST"] = {"NAME": os.path.join(BASE_DIR, "db_test.sqlite3")}
```
natkam: Actually, why do we need this setting?

aqueduct/gateway/tests/utils/mcp.py
```diff
@@ -0,0 +1,192 @@
...
+
+        os.makedirs(os.path.dirname(MCP_CONFIG_PATH), exist_ok=True)
+        with open(MCP_CONFIG_PATH, "w") as f:
+            json.dump(MCP_TEST_CONFIG, f)
+
+    @classmethod
+    def _update_mcp_config_port(cls, port: int):
+        import json
+
+        config = MCP_TEST_CONFIG.copy()
```
natkam: Is `copy` enough? Don't we need a `deepcopy` here? (If we don't need a deepcopy, then I guess we also don't need a copy...)

aqueduct/gateway/tests/test_mcp.py
```diff
@@ -0,0 +1,712 @@
...
+
+from django.conf import settings
+from pydantic.networks import AnyUrl
+
+from gateway.tests.utils.mcp import MCPLiveServerTestCase
+
+if settings.TESTING:
+    logger = logging.getLogger("aqueduct")
+    logging.disable(logging.NOTSET)
+    logger.setLevel(logging.DEBUG)
```
natkam: Instead of adding this piece of code to multiple test modules, maybe we could put it somewhere where it would be executed whenever the tests run? Or even in `settings.py`? Then it would also be easier to disable logging altogether, if you only want to run the tests but don't care about the logs.

aqueduct/gateway/tests/test_mcp.py
```diff
@@ -0,0 +1,712 @@
...
+    async def test_call_tool_long_running(self):
+        """Test calling a long-running tool with progress updates."""
+        async with self.client_session() as session:
+            await session.initialize()
+            tool_name = "longRunningOperation"
+            result = await session.call_tool(tool_name, {"duration": 3, "steps": 5})
+
+            self.assertIsNotNone(result)
+            self.assertIsInstance(result.content, list)
+            self.assertGreater(len(result.content), 0)
```
natkam: Do we test anything related to progress updates here? Is there any way to do it? Otherwise this test doesn't differ much from the `test_call_tool`, and as such seems redundant...?

aqueduct/gateway/views/decorators.py
```diff
@@ -465,3 +523,117 @@ async def wrapper(request: ASGIRequest, *args, **kwargs):
...
+                return JsonResponse({"error": "Mcp-Session-Id header required"}, status=400)
+
+            kwargs["json_rpc_message"] = json_rpc_message
+            kwargs["is_initialize"] = is_initialize
+
+            return await view_func(request, *args, **kwargs)
+        except json.JSONDecodeError as e:
+            log.error(f"JSON decode error: {str(e)}, body was: {request.body!r}")
+            return JsonResponse({"error": f"Invalid JSON: {str(e)}"}, status=400)
+        except Exception as e:
```
natkam: And it doesn't seem to me like this `except Exception` is needed; you check for anything that can go wrong, or catch other exceptions.

aqueduct/management/views/mcp_servers.py
```diff
@@ -0,0 +1,29 @@
...
+        try:
+            mcp_config = get_mcp_config()
+            # Add default icon_url fallback for servers that don't have it
+            for server_name, server_config in mcp_config.items():
+                if "icon_url" not in server_config:
+                    server_config["icon_url"] = "/static/icons/mcp.svg"
+            context["mcp_servers"] = mcp_config
+            # Add the current site's domain for constructing full URLs
+            context["site_host"] = f"{self.request.scheme}://{self.request.get_host()}"
+        except Exception as e:
```
natkam: What sort of exception do you have in mind here? Is there anything that can error out apart from `get_mcp_config()`?

aqueduct/management/admin.py
```diff
@@ -52,11 +62,45 @@ def clean_excluded_models(self):
...
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["excluded_models"].choices = [(m, m) for m in get_model_choices()]
+        if self.instance and getattr(self.instance, "excluded_models", None):
+            self.initial["excluded_models"] = self.instance.excluded_models
+
+        self.fields["excluded_mcp_servers"].choices = [(s, s) for s in get_mcp_server_choices()]
+        if self.instance and getattr(self.instance, "excluded_mcp_servers", None):
+            self.initial["excluded_mcp_servers"] = self.instance.excluded_mcp_servers
```
natkam: This piece of code seems to be duplicated - should already have been called in `ExcludedMCPServersAdminForm.__init__()`, no?

aqueduct/management/admin.py
```diff
@@ -52,11 +62,45 @@ def clean_excluded_models(self):
...
+class UserProfileAdminForm(ExcludedModelsAdminForm, ExcludedMCPServersAdminForm):
     class Meta(ExcludedModelsAdminForm.Meta):
         model = UserProfile
         fields = "__all__"
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["excluded_models"].choices = [(m, m) for m in get_model_choices()]
+        if self.instance and getattr(self.instance, "excluded_models", None):
+            self.initial["excluded_models"] = self.instance.excluded_models
```
natkam: Actually, isn't this piece of code also duplicated - in `ExcludedModelsAdminForm.__init__()`?

pyproject.toml
```diff
@@ -7,13 +7,14 @@ requires-python = ">=3.11"
 dependencies = [
     "asgiref>=3.8.1",
     "celery[redis]>=5.5.2",
-    "daphne>=4.1.2",
-    "django>=5.2",
+    "daphne==4.2.1",
+    "django==5.2.7",
```
natkam: Out of curiosity, why did you pin the version of `daphne` and `django`, when most other packages just set the minimal required version?

aqueduct/gateway/views/mcp.py
```diff
@@ -0,0 +1,724 @@
...
+
+            # Stop the session (this will wait for operations to complete)
+            await session.stop()
+
+            # Now remove from manager after all operations are done
+            async with self._lock:
+                self._sessions.pop(session_id, None)
+            log.info(f"Session {session_id} terminated and removed")
+        else:
+            log.warning(f"Attempt to terminate non-existent session {session_id}")
```
natkam: Is this OK, or should it also raise `SessionNotFoundError`?

natkam (commented):
Sorry it took so long, but this was a huge PR!

docs/user-guide/admin.md
```diff
@@ -56,6 +56,58 @@ Excluded models are not available in any endpoints (returns 404) and are filtere
...
+- **For Service Account Tokens**: Team → Org → Global Settings
+
+The `merge_mcp_server_exclusion_lists` field works identically to `merge_exclusion_lists`:
+- When enabled (default), the exclusion list includes servers from the current level plus all higher levels
+- When disabled, only the current level's exclusions apply, stopping the upward merge
+
+**Example:**  
+A Team excludes `["server-a"]` with merge enabled; its Org excludes `["server-b"]` with merge disabled; global settings exclude `["server-c"]`. Service accounts in that Team would have an effective exclusion list of `["server-a", "server-b"]`—the Org's merge disabled prevents the global `server-c` from being included.
+
+You can configure the global default MCP server exclusion list in `settings.py` using the `AQUEDUCT_DEFAULT_MCP_SERVER_EXCLUSION_LIST` setting (defaults to an empty list).
```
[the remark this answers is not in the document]
meffmadd: It is currently configured in settings.py, similar to the model exclusion list.

aqueduct/gateway/tests/utils/mcp.py
```diff
@@ -0,0 +1,192 @@
...
+        ]
+
+        # Allow test hosts in Django's ALLOWED_HOSTS so requests reach our security decorator
+        # Add common test hosts that we'll use to test the security decorator
+        if "evil.testserver" not in settings.ALLOWED_HOSTS:
+            settings.ALLOWED_HOSTS.append("evil.testserver")
+        if "malicious.com" not in settings.ALLOWED_HOSTS:
+            settings.ALLOWED_HOSTS.append("malicious.com")
+
+        print(
```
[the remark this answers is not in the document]
natkam: If you don't want to confuse the two, you can configure a separate logger for test logs. You can read about Django logging settings in [the documentation](https://docs.djangoproject.com/en/5.2/topics/logging/). I believe something like this should do the job:
```
# settings.py
LOGGING = {
    ...
    "loggers": {
        ...
        "gateway.tests": {
            "handlers": ["console"],
            "level": "DEBUG",
            "propagate": False,
        },
        ...
}
```

aqueduct/gateway/tests/test_mcp.py
natkam: But I'd like to be able to run tests and not have them spit out thousands lines of output that is irrelevant for me, but just show me the errors, if there are any. And now I need to comment out these lines in 2 or 3 files, and then remember not to commit them accidentally. So for me it's a rather inconvenient solution.

By the way, why do MCP tests have to be separated from other tests? What do you mean?

PR #83, opened 2025-12-05
12-responses-api

meffmadd:


aqueduct/gateway/views/decorators.py
```diff
@@ -729,3 +732,81 @@ async def wrapper(request: ASGIRequest, *args, **kwargs):
...
+def check_tool_availability(view_func):
+    @wraps(view_func)
+    async def wrapper(request: ASGIRequest, *args, **kwargs):
+        token: Token | None = kwargs.get("token", None)
+        pydantic_model: ResponseCreateParams | None = kwargs.get("pydantic_model", None)
+        if not token or not pydantic_model:
+            return JsonResponse({"error": "Invalid request"}, status=400)
+
+        tools: Iterable[ToolParam] = pydantic_model.get("tools", None)
+        if tools:
```
natkam: How about:
```suggestion
        tools: Iterable[ToolParam] = pydantic_model.get("tools") or []
```
And then the `if` should not be necessary. There's enough indentation already ;)

aqueduct/gateway/views/decorators.py
```diff
@@ -729,3 +732,81 @@ async def wrapper(request: ASGIRequest, *args, **kwargs):
...
+
+        if response["email"] != token.user.email:
+            return JsonResponse({"error": "Response not found"}, status=404)
+
+        return await view_func(request, response_id, *args, **kwargs)
+
+    return wrapper
+
+
+def check_tool_availability(view_func):
```
natkam: Can you add a docstring that summarises what this decorator does, where it can be used, and why?

aqueduct/gateway/views/responses.py
```diff
@@ -0,0 +1,121 @@
...
+@csrf_exempt
+@require_http_methods(["GET", "DELETE"])
+@token_authenticated(token_auth_only=True)
+@tos_accepted
+@log_request
+@validate_response_id
+@catch_router_exceptions
+async def response(request: ASGIRequest, response_id: str, token: Token, *args, **kwargs):
+    """Combined handler for GET and DELETE /v1/responses/{response_id}"""
+    response = get_response_from_cache(response_id)
```
natkam: Do we need to handle the case when the response is not yet in the cache? `get_response_from_cache` can also return `None`, can't it?

aqueduct/gateway/config.py
```diff
@@ -28,6 +31,8 @@ def _validate_router_config(config: dict):
 def get_router_config() -> dict:
     path = settings.LITELLM_ROUTER_CONFIG_FILE_PATH
     try:
+        log.info(f"Loading router config from {path}")
+        print(f"Loading router config from {path}")
```
natkam: Some debugging leftovers? 😁

natkam:
The Available Endpoints table in the docs index has to be updated too! (and the links list below, I guess)

PR #104, opened 2026-01-29
Batch and Files API rewrite + Vector Stores

meffmadd:


aqueduct/gateway/views/files.py
```diff
@@ -33,115 +63,259 @@ def validate_batch_file(data: bytes):
...
-    file: bytes
-    purpose: Literal["batch", "user_data"]
-    # `File` type requires arbitrary_types_allowed for model settings
-    model_config = ConfigDict(arbitrary_types_allowed=True)
+async def sync_batch_file_if_needed(
+    remote_file_id: Optional[str],
+    token: Token,
+    client: AsyncOpenAI,
+    batch_obj: Optional["Batch"] = None,
+    field_name: str = "output_file",
```
natkam:
```suggestion
    field_name: Literal["output_file", "error_file"] = "output_file",
```

As far as I understand, we only allow two values here: "output_file" or "error_file" - is it correct? We could also add validation to make sure that nobody tries to accidentally call this function with, say, "input_file", or whatever else (but I'm fine with just adding a more specific type annotation).

aqueduct/gateway/views/files.py
```diff
@@ -9,17 +10,46 @@
...
+        return remote_expires_at
+    return local_expires_at
 
 
 def validate_batch_file(data: bytes):
-    lines = data.decode().splitlines()
+    """Validate batch file format: valid JSON lines with unique custom_ids."""
+    lines = data.decode("utf-8").splitlines()
     custom_ids = set()
     for i, line in enumerate(lines):
```
natkam:
```suggestion
    for i, line in enumerate(lines, start=1):
```
... and then just `f"Invalid JSON at line {i}"` - will be more readable this way, I guess :)

aqueduct/management/models.py
```diff
@@ -784,93 +794,12 @@ def model(self) -> openai.types.FileObject:
...
-        try:
-            self.delete_file()
-        except Exception:
-            # Ignore errors deleting the file
-            pass
-        # Then delete the database record
+        """
+        Override ORM delete - files are stored upstream, so we only delete the local DB record.
+        """
         super().delete(using=using, keep_parents=keep_parents)
```
natkam: If I remember correctly, you wanted to change this method still, didn't you? The way it is now, it is redundant.

PR #105, opened 2026-01-29
60: mock test server

natkam:
### High-level description of the changes

I've added a **Django test runner** that starts a **FastAPI mock server** when running any gateway tests - somewhat similar to the mock MCP server which we already use, except all the tests share the same server instance (to avoid the overhead, some port issues, etc). The test runner automatically captures all OpenAI API requests and redirects them to the mock server.

If we want to, it's possible to run the tests against the actual API by setting an env variable `TESTS_USE_MOCK_API` to `False`.

The mock server uses a **default** response for each endpoint, based on:
- the endpoint's path, 
- the request method (but only `responses` require patching `GET` and `DELETE` requests), 
- and possibly `stream=True` sent in the body (for endpoints returning both a JSON and a streaming responses).

Further special mock responses can be configured on a test-basis, using a context manager to wrap the request.

**Note**: The **batches API tests** have already been mocked, i.e. they don't send out any external requests; therefore I haven't yet rewritten them. (Meaning: this is still a TODO.)

The server can also be run as a standalone process from the command line:
```
python -m mock_api.mock_server --port 45999
```
This can be useful e.g. for debugging, and allows setting a specific port or log level (further options to come).

### TODOs:
- Update the batches API tests to use the mock server - existing mocks could be removed then.
- More flexible configuration possibilities for the delays added to the mocked responses (e.g. image generation takes longer than chat completion).
- It'd perhaps be nice to have an option to conveniently disable the mocking on a per-test basis. For now, it can be done by passing the `TESTS_USE_MOCK_API=false` in the command line, and specifying the concrete test to run by name.
- There are still some rough edges, missing docstrings, and a few solutions that perhaps should be improved. Any feedback is welcome!

But the tests pass - so in principle this could be merged, and I can work on the locust tests on a separate branch.

### DIFF
--- a/aqueduct/gateway/tests/test_user_id.py
+++ b/aqueduct/gateway/tests/test_user_id.py
@@ -6,6 +6,7 @@ from urllib.parse import urlparse
 
 import httpx
 from asgiref.sync import sync_to_async
+from django.conf import settings
 from django.core.files.uploadedfile import SimpleUploadedFile
 from django.test import override_settings
 from django.urls import reverse
@@ -22,7 +23,8 @@ ROOT_DIR = Path(__file__).parent.parent.parent.parent
 @override_settings(
     OIDC_OP_JWKS_ENDPOINT="https://example.com/application/o/example/jwks/",
     LITELLM_ROUTER_CONFIG_FILE_PATH=Path(ROOT_DIR / "example_router_config.yaml"),
-    AQUEDUCT_FILES_API_URL="https://api.openai.com",
+    AQUEDUCT_FILES_API_URL="https://files-api.example.com",
+    AQUEDUCT_FILES_API_KEY="test_key",
 )
 class TestUserId(MCPLiveServerTestCase):
     @classmethod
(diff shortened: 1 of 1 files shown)


aqueduct/gateway/tests/test_endpoints.py
```diff
@@ -568,7 +587,7 @@ def test_tika_error_response(self):
         )
         self.assertIn("Tika error extracting text from file", response.json()["error"])
 
-    @override_settings(RELAY_REQUEST_TIMEOUT=0.1)
+    @override_settings(RELAY_REQUEST_TIMEOUT=0.0001)
```
meffmadd: Could this stop working if the tests run faster?

aqueduct/gateway/tests/test_tts_stt.py
```diff
@@ -213,21 +221,30 @@ def test_transcriptions_endpoint_missing_file(self):
...
-            headers=self.multipart_headers,
-        )
-
-        self.assertEqual(
-            response.status_code, 404, f"Expected 404 Not Found, got {response.status_code}"
-        )
+        self.assertIn("TranscriptionCreateParamsStreaming.file", response.json()["error"])
+        self.assertIn("Field required", response.json()["error"])
+
+    # def test_transcriptions_endpoint_non_stt_model(self):
```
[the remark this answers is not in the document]
natkam: I've deleted the commented-out test.

aqueduct/gateway/tests/test_endpoints.py
natkam: I've left it as it is for now. I may come back to it when I make request delays more configurable.

PR #115, opened 2026-03-09
Stricter ruff rules

### LINKED ISSUE #114: Stricter `ruff` rules
Also add other tools:
- `vulture` - Dead Code Detection
- `jscpd` - Duplicate Code Detection
- `radon` - Complexity Metrics

meffmadd:


pyproject.toml
```diff
@@ -33,25 +33,147 @@ dev = [
...
     "fastapi>=0.115.12",
     "pre-commit>=4.2.0",
-    "ruff>=0.13.3",
+    "ruff>=0.15.5",
     "tbump>=6.11.0",
 ]
 
 [tool.ruff]
-line-length = 100
+line-length = 120
```
natkam: What's the reason for this change?

I'm very much in favour of keeping the line length at 100. Otherwise I'll have to scroll the editor window left and right when I happen to work without the huge monitor in the office.

pyproject.toml
```diff
@@ -33,25 +33,147 @@ dev = [
...
+"examples/*.py" = ["PTH", "SIM", "SIM115", "BLE001", "ERA", "T20"]
+"file_id_debug.py" = ["PTH", "SIM", "SIM115", "ERA", "T20"]
+# BLE001: Blind except is intentional in views to return proper error responses
+"**/views/*.py" = ["BLE001"]
+# Phase 3: Additional BLE001 ignores
+"celery.py" = ["BLE001"]
+# E731: Lambda assignments in usage.py are intentional for datetime rounding
+"management/views/usage.py" = ["E731"]
+# S603: Mock API server subprocess call is intentional for test infrastructure
+"mock_api/*.py" = ["S603"]
```
natkam: I can only guess how you arrived at this per-file-ignores list, but it looks to me like it defeats the point of having stricter rules in the first place: it seems to include the majority of the Python files that we have in the project, and makes it virtually impossible to keep track of what rules we actually enforce - and hence to understand what coding standards we want to adhere to. It'll also be impossible to maintain this config, because whenever someone adds a new file, they'll probably also end up adding more rules. The existing ones are already quite redundant and, in many cases, contradictory, and this is only going to get worse with time.

A few examples:

In the `select` list, you enable the whole `"SIM"` group, and then add half a dozen specific `"SIM1xx"` rules. Then you disable most of them on a per-file basis. It would make much more sense to decide which rules we actually want to enforce, and only enable them.   
`"BLE001"` seems to be disabled in most files (admin, views, models, tests, celery, and examples - what else do we even have where a `try/except` could appear?) If you want to disable it, add `"BLE001"` to `ignores`, or remove `"BLE"` from `select`.  
And so on.

Also: If there are specific single cases where it makes sense to violate a rule, it's better to add a `# noqa: E731` comment on a given line rather than configuring a file-based exception (`"management/views/usage.py" = ["E731"]`).

pyproject.toml
```diff
@@ -33,25 +33,147 @@ dev = [
...
+    "PLR0402", # Use `from` instead of `import ... as ...`
+    "PLR1714", # Repeated equality comparisons -> use `in`
+    "PLR2004", # Avoid magic values in comparisons
+    "PLR5501", # Use `elif` instead of `else` then `if`
+    "PLW",     # pylint ruleset
+    "ANN201", # Missing return type annotation for public function
+    "ANN202", # Missing return type annotation for private function
+    "ANN204", # Missing return type annotation for special method
+    "ANN205", # Missing return type annotation for staticmethod
+    "ANN206", # Missing return type annotation for classmethod
```
natkam, line 55: I'm all for using type annotations, but enforcing them without using a type checker (like mypy) to verify that they are correct is likely to end up being counterproductive. We only enforce an arbitrary subset of `"ANNxxx"` rules, but I do not know what the rationale for choosing them was. Do you want to only enforce annotations for returned type? But are these all rules regarding return type? What if a new one is added?

Speaking of annotations, there's also the `"TC"` rule, and `mypy-init-return = true` (in `[tool.ruff.lint.flake8-annotations]`, which would only make sense if we actually used mypy.

Last but not least, I find `"ANN401"` overly strict; we should definitely avoid using `Any` when it's not necessary, but occasionally it *is* the correct type.

pyproject.toml
```diff
@@ -33,25 +33,147 @@ dev = [
...
+
 [tool.ruff.format]
 docstring-code-format = true
 skip-magic-trailing-comma = true
 
 [tool.ruff.lint.isort]
 known-first-party = ["aqueduct", "gateway", "management", "mock_api"]
 
+[tool.ruff.lint.flake8-annotations]
+mypy-init-return = true
```
natkam, line 131: By the way: if we actually want to add mypy, I'd do it in a separate PR. And if we do, we could also start by enforcing type annotations in individual modules, and then expanding to the whole codebase - otherwise it would be a lot of work.

natkam (changes requested):
I think we should discuss these changes on Monday. I would like to know the rationale behind some of them.

pyproject.toml
```diff
@@ -33,25 +33,147 @@ dev = [
...
     "pre-commit>=4.2.0",
-    "ruff>=0.13.3",
+    "ruff>=0.15.5",
     "tbump>=6.11.0",
 ]
 
 [tool.ruff]
-line-length = 100
+line-length = 120
+target-version = "py311"
```
natkam: Looks like this is redundant too: https://docs.astral.sh/ruff/settings/#target-version

If we drop support for 3.11, it may actually end up being confusing, because `target-version` overrides the `requires-python` setting.

natkam (approved):

PR #132, opened 2026-04-09
Add type checker

meffmadd:


aqueduct/gateway/views/vector_store_files.py
```diff
@@ -116,8 +114,9 @@ async def vector_store_files(
         )
 
     # POST /v1/vector_stores/{vector_store_id}/files - Add file to vector store
-    params = pydantic_model if pydantic_model else {}
-    file_id = params.get("file_id")
+    if not pydantic_model:
+        return error_response("Missing required parameter: file_id", param="file_id", status=400)
```
natkam, line 118: This message seems wrong.

aqueduct/gateway/views/utils.py
```diff
@@ -135,16 +136,27 @@ def oai_client_from_body(model: str, request: ASGIRequest) -> tuple[openai.Async
...
+    if deployment is None:
+        log.exception("Model '%s' not found in router deployments", model)
+        raise openai.NotFoundError(
+            message=f"Model '{model}' not found!",
+            response=httpx.Response(
+                request=httpx.Request(method=request.method, url=request.build_absolute_uri()),
+                status_code=404,
+            ),
+            body=None,
+        ) from None
```
natkam, line 150: Is it necessary to raise `from None` here? It's not like there is any other exception to raise from, is there?

aqueduct/gateway/views/mcp.py
```diff
@@ -643,10 +645,12 @@ async def handle_post_request(
 
     session = await session_manager.get_session_with_retry(session_id)
 
-    err_response = _validate_session(session, request_id, name)
+    err_response = _validate_session(session, session_id, name)
     if err_response:
         return err_response
 
+    assert session is not None  # noqa: S101
```
natkam: Hm. If this linting rule clashes with mypy, maybe we should disable it altogether? `S101` doesn't seem to be very important to me.

aqueduct/gateway/views/mcp.py
```diff
@@ -491,10 +493,10 @@ def parse_session_message(
     reqeust_id: str | int,
     session_id: str,
     json: bool = False,
-) -> dict | str:
+) -> dict[str, Any] | str:
     if isinstance(received_message, Exception):
         log.error("Session %s returned exception: %s", session_id, received_message)
-        jsonrpc_message = JSONRPCError(
+        jsonrpc_message: JSONRPCError | JSONRPCMessage = JSONRPCError(
```
natkam: In this line, `jsonrpc_message` is clearly a `JSONRPCError`, so I actually found this annotation a bit confusing. Would it please mypy if you declared `jsonrpc_message: JSONRPCError | JSONRPCMessage` above the `if`?

aqueduct/gateway/views/mcp.py
```diff
@@ -207,29 +207,29 @@ async def start(self):
...
-                self._httpx_client,  # type: ignore[arg-type]
-                self._write_stream_reader,  # type: ignore[arg-type]
-                self._read_stream_writer,  # type: ignore[arg-type]
+                self._httpx_client,
+                self._write_stream_reader,
+                self._read_stream_writer,
                 self.write_stream,
                 start_get_stream_task,
-                self._task_group,  # type: ignore[arg-type] - Pass our simple task group
+                self._task_group,  # type: ignore[arg-type]
```
natkam, line 224: Actually, can we make `SimpleTaskGroup` inherit from the abstract `TaskGroup`, which `post_writer` expects? Or do they differ too much? (Just a question - I don't know if you've already tried this.)

aqueduct/gateway/views/image_generation.py
```diff
@@ -1,10 +1,11 @@
 import logging
+from typing import Any
 
 from django.core.handlers.asgi import ASGIRequest
 from django.http import JsonResponse
 from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import require_POST
-from litellm import BadRequestError
+from litellm import BadRequestError  # type: ignore[attr-defined]
```
natkam: Looks like `litellm` causes `[attr-defined]` quite a lot. Should we maybe add a general rule for this module in the mypy config?

aqueduct/gateway/views/files.py
```diff
@@ -140,43 +141,59 @@ async def files(
...
+                await sync_to_async(list)(  # type: ignore[call-arg]
+                    FileObject.objects.filter(token__user=token.user)
+                    .order_by("-created_at")
+                    .select_related("token")
+                ),
             )
 
+        serialized_files: list[dict[str, Any]] = [
+            file.model.model_dump(exclude_none=True, exclude_unset=True) for file in file_objects
+        ]
```
natkam:
```suggestion
        if token.service_account:
            qs = (
                FileObject.objects.filter(token__service_account__team=token.service_account.team)
                .order_by("-created_at")
                .select_related("token")
            )
        else:
            qs = (
                FileObject.objects.filter(token__user=token.user)
                .order_by("-created_at")
                .select_related("token")
            )
        file_objects = [
            f_obj.model.model_dump(exclude_none=True, exclude_unset=True) async for f_obj in qs
        ]

```
Using `async for` seems to please mypy and fix the issue with the unexpected argument to `list`. It's also less verbose, since we end up iterating over the file objects anyway.

For reference: 
- https://docs.djangoproject.com/en/5.2/topics/async/#queries-the-orm   
- https://docs.djangoproject.com/en/5.2/topics/db/queries/#query-iteration

aqueduct/gateway/views/batches.py
```diff
@@ -46,7 +48,7 @@ async def batches(
         else:
             batch_qs = Batch.objects.filter(token__user=token.user)
 
-        batch_objects = await sync_to_async(list)(
+        batch_objects: list[Batch] = await sync_to_async(list)(  # type: ignore[call-arg]
```
natkam: Same fix should be applicable here!

aqueduct/gateway/views/__init__.py
```diff
@@ -1,18 +1,59 @@
...
+from .vector_store_file_batches import (
+    vector_store_file_batch_files as vector_store_file_batch_files,
+)
+from .vector_store_file_batches import vector_store_file_batches as vector_store_file_batches
+from .vector_store_files import vector_store_file as vector_store_file
+from .vector_store_files import vector_store_file_content as vector_store_file_content
+from .vector_store_files import vector_store_files as vector_store_files
+from .vector_stores import vector_store as vector_store
+from .vector_stores import vector_store_search as vector_store_search
+from .vector_stores import vector_stores as vector_stores
```
natkam: I don't think importing stuff `as` itself is necessary if you list this stuff in `__all__`...?

aqueduct/aqueduct/celery.py
```diff
@@ -23,28 +24,28 @@
 app.autodiscover_tasks()
 
 
-@app.task(bind=True, ignore_result=True)
-def delete_old_requests(self):
+@app.task(bind=True, ignore_result=True)  # type: ignore[untyped-decorator]
```
natkam: We could also add
```
# mypy: disable-error-code="untyped-decorator"  # Celery `task` decorators are untyped
```
at the top of the file to disable all three cases. Or, maybe even better, install [`celery-types`](https://pypi.org/project/celery-types/). Just installing the package may be enough.

aqueduct/gateway/authentication.py
```diff
@@ -96,7 +102,7 @@ def authenticate(self, request, **kwargs) -> User | None:
         # E.g. in middleware: request.auth_token = token_instance
         return token_instance.user
 
-    def get_user(self, user_id) -> User | None:
+    def get_user(self, user_id: int) -> AbstractBaseUser | None:  # type: ignore[override]
```
natkam: `AbstractUser` is not a correct type annotation here. If we add an `ignore` comment anyway, I'd rather we kept the return type `User`, which is more useful for us, even if it's not treated as valid annotation by mypy.

```suggestion
    def get_user(self, user_id: int) -> User | None:  # type: ignore[valid-type]
```

aqueduct/gateway/authentication.py
```diff
@@ -38,7 +40,9 @@ class TokenAuthenticationBackend(BaseBackend):
     using the aqueduct.management.models.Token model.
     """
 
-    def authenticate(self, request, **kwargs) -> User | None:
+    def authenticate(  # type: ignore[override]
```
natkam: (same here)

aqueduct/gateway/authentication.py
natkam: Or, what also seems to work:
```
if TYPE_CHECKING:
    from django.contrib.auth.models import User as AuthUser
```

and 
```
    def get_user(self, user_id: int) -> "AuthUser | None":
        ...
```

natkam (approved):

PR #134, opened 2026-04-15
Match MCP servers based on `url` instead of `server_label`

meffmadd:


aqueduct/gateway/views/decorators.py
```diff
@@ -808,30 +814,33 @@ async def wrapper(
 async def _validate_mcp_tool(
     request: ASGIRequest, token: Token, tool: ToolParam
 ) -> ViewResult | None:
-    # Note: mypy doesn't recognise the types of `ToolParam` attributes correctly
```
natkam, line 811: I think it's better to leave a comment explaining why we need those `type: ignore[...]` comments all over this function. It's good practice, according to [Mypy docs](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#when-you-re-puzzled-or-when-things-are-complicated).
But maybe you can think of a better place to put it?

natkam (approved):

PR #135, opened 2026-04-16
Create teams from oauth groups

### LINKED ISSUE #112: Automatically create groups (teams) from OAuth groups
Team creation currently happens manually. We should to this via OAuth groups.

Basically add implementation from Open WebUI: https://docs.openwebui.com/features/access-security/rbac/groups/

It uses the `ENABLE_OAUTH_GROUP_MANAGEMENT ` to enable user assignment to groups from OAuth. And `ENABLE_OAUTH_GROUP_CREATION ` controls if groups (teams) should be created automatically or if users should just get assigned to groups (teams) if they exist already.

We should also implement a configurable filter function for the groups that are eligible for group creation. For example at TU each internal organization starts with "E" followed by a code, so we could filter for these. And there are sub-groups so we could also decide if we keep that.

meffmadd:
Fixes #112


aqueduct/management/migrations/0009_team_oauth_group_name_alter_request_model.py
```diff
@@ -0,0 +1,23 @@
...
+            model_name='team',
+            name='oauth_group_name',
+            field=models.CharField(blank=True, db_index=True, default='', help_text='The OAuth group that created this team (auto-managed, not user-editable)', max_length=255, verbose_name='OAuth group name'),
+        ),
+        migrations.AlterField(
+            model_name='request',
+            name='model',
+            field=models.CharField(blank=True, default='', help_text='Model used in request', max_length=255),
+        ),
+    ]
```
natkam, line 27: Since we're (potentially) changing the content of the `model` column in the database here, it might make sense to run a data migration that automatically replaces any existing `None`/`NULL` values with `""` before applying the migration that alters the `model` field and disallows `None`. If I'm not mistaken, with a PostgreSQL database, this migration would fail with an integrity error if there already were any rows with `NULL`.

Something like this should do (disclaimer: I haven't tested it):
```
def replace_nulls(apps, schema_editor):
    Request = apps.get_model('management', 'Request')
    Request.objects.filter(model__isnull=True).update(model='')

...
    operations = [
        ...
        migrations.RunPython(replace_nulls, migrations.RunPython.noop),
        migrations.AlterField(
            model_name='request',
            name='model',
            field=models.CharField(blank=True, default='', help_text='Model used in request', max_length=255),
        ),
    ]
```

aqueduct/aqueduct/settings.py
```diff
@@ -125,6 +125,45 @@ def my_org_name_extractor(groups: list[str]) -> str | None:
...
+            return None
+    """
+    return None
+
+
+def my_transform(group: str, groups: list[str] | None = None) -> tuple[str, str] | None:
+    if group.startswith("E"):
+        team_name = group.split("-")[0]
+        return (team_name, group)
+    return None
```
natkam: This function's name is rather unspecific. Maybe it'd be better to call it e.g. `oauth_team_names_from_groups`?

Also: it doesn't look like there's a way to use the `default_oauth_team_names_from_groups`, is there? How could users override this setting, actually - without changing the `settings.py` file?

aqueduct/management/tests/test_sync_oauth_team_names_admin.py
```diff
@@ -0,0 +1,176 @@
...
+        Team.objects.create(name="E123", org=self.org, oauth_group_name="E123-Students")
+
+        request = self._create_request()
+        queryset = Team.objects.all()
+
+        with override_settings(OAUTH_TEAM_NAMES_FROM_GROUPS_FUNCTION=team_names_empty):
+            sync_oauth_team_names_action(self.modeladmin, request, queryset)
+
+        team = Team.objects.get(oauth_group_name="E123-Students")
+        self.assertIsNotNone(team)
```
natkam, line 131: The docstring claims that there's a warning message tested in this test - where, though? Are there some assertions about it that could be added, maybe?

aqueduct/management/admin.py
```diff
@@ -174,6 +179,114 @@ def delete_tos_cache(modeladmin, request, queryset):
...
+            log.info(
+                "Updated team '%s' → '%s' (org: %s, oauth_group: '%s')",
+                old_name,
+                new_name,
+                team.org.name,
+                team.oauth_group_name,
+            )
+            updated_count += 1
+
+        skipped_count = len(teams_skipped)
```
natkam: Is it important to update the teams one-by-one here? This could be achieved with [`bulk_update()`](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#bulk-update) in one query, without the need to manually start a transaction. (Unless we have to call the `save()` method for each team, which `bulk_update` doesn't do.)

aqueduct/management/admin.py
```diff
@@ -174,6 +179,114 @@ def delete_tos_cache(modeladmin, request, queryset):
...
+    """
+    func = getattr(settings, "OAUTH_TEAM_NAMES_FROM_GROUPS_FUNCTION", None)
+    if not func:
+        modeladmin.message_user(
+            request, "OAUTH_TEAM_NAMES_FROM_GROUPS_FUNCTION not configured", level=messages.WARNING
+        )
+        return
+
+    # Only process OAuth-managed teams
+    oauth_teams = queryset.filter(oauth_group_name__gt="").select_related("org")
```
natkam: Do you want to exclude the teams with the `oauth_group_name == ""`? If yes, you can use:

```suggestion
    oauth_teams = queryset.exclude(oauth_group_name="")
```

Also, if I'm not mistaken, there's no need to select the related `org`, since we're not using these objects anywhere (filtering by `org=team.org` will just use the foreign key, so it doesn't need the whole object).

aqueduct/management/admin.py
```diff
@@ -174,6 +179,114 @@ def delete_tos_cache(modeladmin, request, queryset):
...
+    teams_to_delete = []
+    teams_skipped = []
+
+    all_oauth_groups = list(
+        Team.objects.filter(oauth_group_name__gt="")
+        .values_list("oauth_group_name", flat=True)
+        .distinct()
+    )
+
+    tuple_length = 2
```
natkam: This variable name doesn't say much about what the variable actually represents. What tuple are we talking about? Why should it equal 2? It is only used in one place anyway, so I'd be in favour of skipping its definition altogether: it doesn't make the "magic number" any less magical, so to speak.

aqueduct/management/auth.py
```diff
@@ -41,6 +42,120 @@ def _org(self, groups: list[str]) -> Org | None:
...
+                result = func(group, groups)
+                if not isinstance(result, tuple) or len(result) != 2:  # noqa: PLR2004
+                    continue
+                team_name, original_name = result
+                if not (team_name and isinstance(team_name, str) and original_name):
+                    continue
+                team_mappings.append((team_name.strip(), original_name.strip()))
+        except Exception as e:
+            log.exception("Error calling OAUTH_TEAM_NAMES_FROM_GROUPS_FUNCTION: %s", e)
+            return []
```
natkam: If we actually want to catch errors calling `OAUTH_TEAM_NAMES_FROM_GROUPS_FUNCTION`, then I'd say only the line where the function is called should be in the `try` block. Otherwise we're catching errors in our code, which should be unnecessary.

aqueduct/management/auth.py
```diff
@@ -41,6 +42,120 @@ def _org(self, groups: list[str]) -> Org | None:
...
+                    if not enable_creation:
+                        log.info(
+                            "Skipping team creation for '%s' (ENABLE_OAUTH_GROUP_CREATION=False)",
+                            team_name,
+                        )
+                        continue
+                    team = Team.objects.create(
+                        name=team_name, org=org, oauth_group_name=original_group_name
+                    )
+                    log.info("Created team '%s' for org '%s'", team_name, org.name)
```
natkam: Question: If a team does not exist, but creation is disabled, do we care?

Because if we don't, this looks like a usecase for [`get_or_create()`](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#get-or-create)!
```suggestion
                if not enable_creation:  # this could even be outside of the loop - no need to send db queries for teams if we don't use them
                    log.info(
                        "Skipping team '%s' (ENABLE_OAUTH_GROUP_CREATION=False)",
                        team_name,
                    )
                    continue
                team, created = Team.objects.get_or_create(name=team_name, org=org, defaults={"oauth_group_name": original_group_name})
                if created:
                    log.info("Created team '%s' for org '%s'", team_name, org.name)
                else:
                    log.info("Reused existing team '%s' for org '%s'", team_name, org.name)
```

aqueduct/management/tests/test_oauth_team_creation.py
```diff
@@ -0,0 +1,516 @@
...
+class OAuthTeamSettingsTestCase(TestCase):
+    """Test OAuth team settings behavior."""
+
+    def setUp(self):
+        self.backend = OIDCBackend()
+        self.org = Org.objects.create(name="test-org")
+        self.user_group, _ = Group.objects.get_or_create(name="user")
+
+    def test_default_function_returns_none(self):
+        """Test that default function returns None (no teams)."""
```
natkam: Wait, but isn't this just a test that `lambda group, groups=None: None` returns `None`? It doesn't have anything to do with our code in the project. It only tests the test setup, and that `override_settings` works properly (which is not our responsibility to test).

natkam (changes requested):
Bottom line: How could users override the `OAUTH_TEAM_NAMES_FROM_GROUPS_FUNCTION` setting, actually - without changing the settings.py file?

aqueduct/aqueduct/settings.py
meffmadd: The `my_transform` function was a mistake and should not have been committed. Callables have to be configured by updating `settings.py`. See #143

aqueduct/management/admin.py
meffmadd: It has to be defined by the ruff magic value rule. Of course it is still a magic value so that is one case where the rule does not really help.

natkam (approved):

natkam (approved):

PR #136, opened 2026-04-16
Use built-in Django async queryset methods instead of `sync_to_async`

natkam:
I've noticed that we used the same trick with `sync_to_async(list)(...<query>...)` in other places, and also realised that some of the `sync_to_async` calls on Queryset methods are unnecessary, because Django has asynchronous versions of those methods.


meffmadd (approved):
LGTM!

Issue #102, opened 2026-01-19
Add "token balances" for usage limiting


meffmadd:
Inspired by: https://www.librechat.ai/docs/configuration/librechat_yaml/object_structure/balance

PR #139, opened 2026-04-23
Apply request limits on a weighted per-model basis

### LINKED ISSUE #137: Add per model request limits
Smaller models should naturally have a higher request limit.

Maybe add a "multiplier" to `router_config.yaml` to multiply request limits based on the base request limit based on `Token.get_limit()`. A small model could get a multiplier of e.g. 2 meaning the limit is doubled for this model.

The multiplier should be visible in the "Models" UI page. (Maybe with a tag if multiplier != 1 like "1.5x Limits" or something)

meffmadd:
This PR introduced weighted request limits on a per model basis. So if a model has "2x Limits" each request to the model counts as half towards the request limit. So the limits become semantically more like a budget. This is highly related to #102


aqueduct/gateway/views/decorators.py
natkam: Just caught my eye ;)

```suggestion
                if exceeded:
```
[PEP8 recommends](https://pep8.org/#programming-recommendations): "For sequences, (strings, lists, tuples), use the fact that empty sequences are false".

aqueduct/gateway/views/decorators.py
```diff
@@ -278,33 +279,60 @@ async def wrapper(request: ASGIRequest, *args: Any, **kwargs: Any) -> ViewResult
                 # Define the time window for usage check (last 60 seconds)
                 time_window_start = timezone.now() - timedelta(seconds=60)
 
+                # Build query filter - no model filter (count all requests)
+                query_filter: dict[str, Any] = {"token": token, "timestamp__gte": time_window_start}
```
natkam: Nit-picking, but the code below would be easier to understand if the name of this variable reflected what sort of filtering we were doing here. Or, maybe even better, define a queryset that can then be used as a base for aggregations and other processing:

```suggestion
                recent_requests: Queryset[Request] = Request.objects.filter(token=token, timestamp__gte=time_window_start)
```
The queryset won't be evaluated here, so no additional db queries will be performed.

aqueduct/gateway/views/decorators.py
```diff
@@ -278,33 +279,60 @@ async def wrapper(request: ASGIRequest, *args: Any, **kwargs: Any) -> ViewResult
...
+                # Get per-model request counts for weighted budget calculation
+                model_counts: dict[str, int] = {}
+                async for item in (
+                    Request.objects.filter(**query_filter)
+                    .values("model")
+                    .annotate(request_count=Count("id"))
+                ):
+                    model = item.get("model")
+                    if model:
+                        model_counts[model] = item["request_count"]
```
natkam: If you add `.filter(model__isnull=False)` to the query, the `if model:` check shouldn't be necessary :)

And you could even do something like:

```suggestion
                model_requests: dict[str, int] = {
                    req["model"]: req["request_count"] async for req in (
                        Request.objects.filter(**query_filter)  # or: recent_requests
                        .filter(model__isnull=False)
                        .values("model")
                        .annotate(request_count=Count("id"))
                    )
                }
```

Note: if the migration changing the default empty value for `model` is applied first, instead of filtering out nulls, we'll need `.exclude(model="")`.

natkam (approved):
nit-picks only (;

natkam (approved):

Issue #141, opened 2026-05-04
Add token generation endpoint


meffmadd:
**Is your feature request related to a problem? Please describe.**
Aqueduct could be used programmatically by applications if token generation was possible through an API.

**Describe the solution you'd like**
To generate a token we need 3 pieces of information:

1. token name
2. user email (tokens are always associated with a user)
3. org name (**should be automatically determined from the user who made the request**)
4. expiration data (optional)

For the **plain endpoint** we have to send this info along with it:
``` bash
curl -X PUT -H "Authorization: Bearer <user_token>" \
https://aqueduct.example.com/aqueduct/management/tokens/generate \
-H "Content-Type: application/json" \
-d '{
    "name": "new-token",
    "user": "other@example.com",
    "expires": "2026-05-11Z10:00:00.000"
}'
```

For the **JWT endpoint** we have the JWT in the body and the user information is read from there (fails if other user would map to a different org):
``` bash
curl -X PUT -H "Authorization: Bearer <user_token>" \
https://aqueduct.example.com/aqueduct/management/tokens/generate \
-H "Content-Type: application/json" \
-d '{
    "name": "new-token",
    "jwt": "...",
}'
```
This functionality could need a specific role (e.g. `token-write`). See #106

PR #142, opened 2026-05-08
60 locust load tests

### LINKED ISSUE #60: Add `locust` load tests
Create `locust` load tests for the API and management endpoints. API gateway endpoints (models, MCP servers) should be mocked.

The mock should be a file in `gateway/tests/utils` that runs a FastAPI server if it is executed directly but also expose the mock functions as simple python functions that can be imported by tests that need to mock API return values. Might be overkill but it would also be nice to have a validator script that checks if the keys from the actual output of the OpenAI API matches the mock responses. This could be a simple cache proxy server that runs for all tests (read `example_router_config.yaml`, update URL to cache proxy and mock `get_router_config` globally) that intercepts all traffic, has a JSON file with cached requests/responses. If the request was not cached it calls the OpenAI API and caches it, otherwise it just returns the cached response. It should also be possible to refresh the cache and disable it as caching is one of the hard things in computer science :) Streaming calls will be hard to implement tho... In general it would be SO nice if writing/debugging/rewriting tests would not be SO painful.

To run the `locust` load tests, you simply spawn the mock server, spawn Aqueduct and use the local Aqueduct URL in `locust`. One problem will be to create the token automatically or hardcode a token to be able to run the load tests automatically e.g. in CI.

Load tests should run with docker compose setup to test interaction with Postgres db as we often had "remaining connection slots are reserved for non-replication superuser connections" errors there. So mock server should delay requests as well to simulate longer running requests. Instant replies tests more the web server bottleneck, delayed requests in mock server tests more the overall behavior of the system under load. Both should be possible (CLI arg for mock server).

This might be complicated but it would be VERY interesting to see a (Grafana) dashboard of the actual requests ending up at the mock server! This would make testing the settings for request limits under actual load much easier. I guess this is profiling and in this context I found:
- [New Relic](https://docs.newrelic.com/docs/apm/agents/python-agent/async-instrumentation/django-asgi-mode/)
- [Sentry](https://sentry.io/welcome/)

Image

natkam:
For a summary of changes, please see the commit list.

_Note regarding the `test_vector_stores_api.py`: I just split the single `TestCase` class into three, grouping the tests for the vector stores, VS files, and VS file batches, respectively. Since I had to rearrange the order of the tests, Git shows a huge diff. Sorry about that._

How to run the load tests (locally):

0. Install test dependencies:  
  ```
  uv pip install --group dev-locust
  ```
1. `cd aqueduct/locustfiles`
2. Start the application and the mock API server:
```
docker compose -f docker-compose-locust.yaml up --build
```
The Django app runs at `localhost:8000`. The mock server is available at `localhost:45999`.
3. Start locust:
```
locust -f locustfile.py
```
The web interface is available in the browser at `localhost:8089`. A test can be configured and started using the interface.

`locustfile.py` defines a number of tasks that test the "lifecycle" of a given type of resource (file, batch, response, vector store, etc.). 

Each test user has its own Django user created in the database, together with a profile and token. This user-specific data is deleted from the database when a test user is deleted (e.g. by decreasing the number of users, or by stopping the test run). I created endpoints for creation and deletion of test users, since loading them from a fixture is not feasible with a large (and in principle unlimited) number of test users. Also, test users cannot share the same token, because this would cause conflicts when different test users try to modify a resource, or problems with limits of e.g. vector stores or batches allowed per user.

TODOs (based on the issue description):
 - running load tests in CI
 - more granular delays settings for different endpoints (didn't seem to be a priority, but would make tests more realistic)
 - profiling
 - caching of actual OpenAI responses - while it definitely sounds cool, it also seems like an overkill indeed. But let's discuss how we could handle verifying that the mocked responses are accurate.

### DIFF
--- a/aqueduct/gateway/tests/test_vector_stores_api.py
+++ b/aqueduct/gateway/tests/test_vector_stores_api.py
@@ -20,7 +20,7 @@ from management.models import (
 from management.models import VectorStore as VectorStoreModel
 from management.models import VectorStoreFile as VectorStoreFileModel
 from management.models import VectorStoreFileBatch as VectorStoreFileBatchModel
-from mock_api.mock_configs import MockConfig
+from mock_api.mock_configs import MockConfig, vector_store_file_detail_response
 
 
 @override_settings(MAX_USER_VECTOR_STORES=3, MAX_TEAM_VECTOR_STORES=10, MAX_VECTOR_STORE_FILES=100)
@@ -391,12 +391,16 @@ class TestVectorStoreFiles(TestVectorStoresBase):
         file_obj = self._create_file_object()
 
         # Add file to vector store
-        resp = self.client.post(
-            self.vs_files_url,
-            data=json.dumps({"file_id": file_obj.id}),
-            headers=self.headers,
-            content_type="application/json",
+        new_vsf_resp = MockConfig(
+            response_data=vector_store_file_detail_response((self.vs_id, "vsf-mock-123"))
         )
+        with self.mock_server.patch_external_api(self.vs_files_url, new_vsf_resp):
+            resp = self.client.post(
+                self.vs_files_url,
+                data=json.dumps({"file_id": file_obj.id}),
+                headers=self.headers,
+                content_type="application/json",
+            )
         self.assertEqual(resp.status_code, 200, f"Add file failed: {resp.json()}")
         data = resp.json()
         # ID is now the upstream ID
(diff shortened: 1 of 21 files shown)


natkam:
@meffmadd I've adjusted the Dockerfiles, and fixed some small issues. This PR is ready now :)

aqueduct/gateway/tests/test_vector_stores_api.py
meffmadd, line 1: Why were the vector store tests refactored? Was this "just because" or was there some change required for the locust?

aqueduct/management/views/test_auth.py
meffmadd, line 1: Could we rename this file to something like `user.py` in anticipation for the #141 issue where we will have a user + token generation endpoint. This would be the draft implementation of that.

meffmadd (approved):
Looks good! Let's do the file rename in a separate PR.

aqueduct/gateway/tests/test_vector_stores_api.py
natkam: Just because it was hard to find specific tests (or whether a test exists for a certain case) in this one big class, and I was annoyed by it at some point ;) It wasn't strictly necessary.

PR #155, opened 2026-06-08
Replace undersores with hyphens in dc services and container names

natkam:
Hyphen seems to be the preferred separator, and underscore caused DNS issues sometimes (technically, it's not allowed in DNS hostnames).


meffmadd (approved):
LGTM!

PR #156, opened 2026-06-09
UI Updates

### LINKED ISSUE #85: In admin panel, allow for quick filtering of users that have request/token limit set
The admin UI is currently not very intuitive for this. I want an easy way to manage limits for users in the admin UI. maybe even for orgs and teams?

meffmadd:
Closes #85

### DIFF
--- a/aqueduct/management/admin.py
+++ b/aqueduct/management/admin.py
@@ -325,17 +325,17 @@ class UserAdmin(BaseUserAdmin):
     def request_limit(self, obj) -> int:
         return obj.profile.requests_per_minute
 
-    request_limit.short_description = "Requests per minute"
+    request_limit.short_description = "Requests/min (per API key)"
 
     def input_limit(self, obj) -> int:
         return obj.profile.input_tokens_per_minute
 
-    input_limit.short_description = "Input tokens per minute"
+    input_limit.short_description = "Input tokens/min (per API key)"
 
     def output_limit(self, obj) -> int:
         return obj.profile.output_tokens_per_minute
 
-    output_limit.short_description = "Output tokens per minute"
+    output_limit.short_description = "Output tokens/min (per API key)"
 
 
 # Re-register UserAdmin
--- a/aqueduct/management/admin.py
+++ b/aqueduct/management/admin.py
@@ -303,6 +303,28 @@ def reload_from_upstream(modeladmin, request, queryset):
         messages.success(request, f"Successfully reloaded {len(objects)} object(s).")
 
 
+class HasLimitSetFilter(admin.SimpleListFilter):
+    title = "has limit set"
+    parameter_name = "has_limit"
+
+    def lookups(self, request, model_admin) -> list[tuple[str, str]]:
+        return (("yes", "Yes"), ("no", "No"))
+
+    def queryset(self, request, queryset: QuerySet) -> QuerySet:
+        from django.db.models import Q
+
+        has_any = Q(
+            Q(profile__requests_per_minute__isnull=False)
+            | Q(profile__input_tokens_per_minute__isnull=False)
+            | Q(profile__output_tokens_per_minute__isnull=False)
+        )
+        if self.value() == "yes":
+            return queryset.filter(has_any)
+        if self.value() == "no":
+            return queryset.exclude(has_any)
+        return queryset
+
+
 # Define a new User admin
 class UserAdmin(BaseUserAdmin):
     inlines: ClassVar[tuple] = (ProfileInline,)
@@ -315,6 +337,7 @@ class UserAdmin(BaseUserAdmin):
         "output_limit",
     )
     list_select_related: ClassVar[list] = ["profile"]
+    list_filter: ClassVar[list] = ["is_staff", HasLimitSetFilter]
     actions: ClassVar[list] = [make_admin, make_org_admin, make_user, delete_tos_cache]
 
     def get_groups(self, obj) -> str:
--- a/aqueduct/management/admin.py
+++ b/aqueduct/management/admin.py
@@ -9,7 +9,7 @@ from django.conf import settings
 from django.contrib import admin, messages
 from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
 from django.contrib.auth.models import Group, User
-from django.db.models import QuerySet
+from django.db.models import Q, QuerySet
 from django.urls import reverse
 from django.utils.html import format_html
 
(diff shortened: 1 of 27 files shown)


aqueduct/management/templates/management/batches.html
```diff
@@ -137,6 +138,23 @@ <h3 class="font-bold">Preview: {{ batch.error_file_id_preview }}</h3>
...
+                                        {{ batch.expires_dt|date:"Y-m-d H:i" }}
+                                    </span>
+                                {% elif batch.expiration_status == "active" %}
+                                    <span class="badge badge-sm badge-soft badge-success gap-1">
+                                        {{ batch.expires_dt|date:"Y-m-d H:i" }}
+                                    </span>
+                                {% else %}
+                                    <span class="badge badge-sm badge-soft badge-ghost gap-1">Never</span>
+                                {% endif %}
+                            </td>
```
natkam: How about defining a [`partial`](https://docs.djangoproject.com/en/6.1/ref/templates/language/#template-partials) and using it here and in `files.html`? These two pieces of template code seem to be exactly the same :)


```suggestion
                            {% with obj=batch %}
                              {% partialdef expiration-status inline %}
                              <td>
                                  {% if obj.expiration_status == "expired" %}
                                    <span class="badge badge-sm badge-soft badge-error gap-1">
                                      {{ obj.expires_dt|date:"Y-m-d H:i" }}
                                    </span>
                                  {% elif obj.expiration_status == "expiring_soon" %}
                                    <span class="badge badge-sm badge-soft badge-warning gap-1">
                                      {{ obj.expires_dt|date:"Y-m-d H:i" }}
                                    </span>
                                  {% elif obj.expiration_status == "active" %}
                                    <span class="badge badge-sm badge-soft badge-success gap-1">
                                      {{ obj.expires_dt|date:"Y-m-d H:i" }}
                                    </span>
                                  {% else %}
                                    <span class="badge badge-sm badge-soft badge-ghost gap-1">Never</span>
                                  {% endif %}
                              </td>
                              {% endpartialdef %}
                            {% endwith %}
```

aqueduct/management/templates/management/team.html
```diff
@@ -50,7 +56,7 @@ <h2 class="text-xl font-semibold mb-2">Service Accounts</h2>
                                 Owner: {{ sa.token.user.email|default:"(No Owner)" }}
                             </span>
                             <span class="text-sm opacity-50 flex-shrink-0 ml-2">
-                                Token: <span class="font-mono text-xs">{{ sa.token.key_preview }}</span> {# Show preview #}
+                                Key: <span class="font-mono text-xs">{{ sa.token.key_preview }}</span> {# Show preview #}
```
natkam: I don't know if this is on purpose, but `Key` here is inconsistent with `API Key` used elsewhere. I'm leaving a comment just in case it was an oversight.

aqueduct/management/views/models.py
```diff
@@ -27,4 +27,5 @@ def get_context_data(self, **kwargs) -> dict[str, object]:
 
         context["title"] = "Models"
```
natkam: Looks like "title" is not used in the template any more.

aqueduct/management/admin.py
```diff
@@ -303,6 +303,28 @@ async def reload_obj(obj) -> None:
...
 
+class HasLimitSetFilter(admin.SimpleListFilter):
+    title = "has limit set"
+    parameter_name = "has_limit"
+
+    def lookups(self, request, model_admin) -> list[tuple[str, str]]:
+        return (("yes", "Yes"), ("no", "No"))
+
+    def queryset(self, request, queryset: QuerySet) -> QuerySet:
+        from django.db.models import Q
```
natkam: Please move the import to the top of the file :)

aqueduct/management/templates/management/usage.html
```diff
@@ -154,13 +153,24 @@ <h2 class="text-xl font-semibold mb-2">Request Counts by
                                         {% if item.user_is_token_owner %}
                                             <a href="?org={{ selected_org.id }}&span={{ selected_span }}&token={{ item.token_id }}"
                                                class="font-bold link link-hover text-primary">
-                                                {{ item.name|default:"N/A" }}
+                                                {% if item.service_account_name %}
+                                                    {{ item.service_account_name }}
+                                                {% else %}
+                                                    {{ item.name|default:"N/A" }}
+                                                {% endif %}
```
natkam: You could perhaps chain `default:` filters like this:

```suggestion
                                                {% item.service_account_name|default:item.name|default:"N/A" }}
```
It should work in principle, as long as the variable `item` is defined and has the `name` attribute defined (which I believe should be the case; in the view, we always annotate the queryset objects with `name`).

Curiously, if `item` does not have the attribute `service_account_name`, there will be no error - `item.service_account_name` is just going to be treated as a falsey value, and the value from the `default` filter will be used. However, already in the filter, if the `name` attribute does not exist on `item`, Django's going to throw a `VariableDoesNotExist` error.

aqueduct/management/templates/management/usage.html
```diff
@@ -154,13 +153,24 @@ <h2 class="text-xl font-semibold mb-2">Request Counts by
...
                                             </a>
                                         {%  else %}
                                             <span class="font-bold">
-                                                {{ item.name|default:"N/A" }}
+                                                {% if item.service_account_name %}
+                                                    {{ item.service_account_name }}
+                                                {% else %}
+                                                    {{ item.name|default:"N/A" }}
+                                                {% endif %}
```
natkam: Same here: you could probably chain the `default` filters.

natkam (approved):
A few minor suggestions, but in general I think the UI will be more intuitive now 👍

aqueduct/management/templates/management/batches.html
meffmadd: As far as I can tell `partialdef` was introduced in Django 6.0 so this would not work: https://djangotemplatetagsandfilters.com/tags/partialdef/

aqueduct/management/templates/management/team.html
meffmadd: True this should be API Key!

aqueduct/management/templates/management/batches.html
natkam: Ah, you're right. I've tested it locally and it worked, so I haven't thought about this. But I had Django 6.0.6 installed, as we only specify the minimum package version, without an upper limit, and `uv` automatically decided for the highest matching one.

But (as a side question) is there anything stopping us from moving to Django 6? It's not like `aqueduct` is a library used as a dependency with other packages, where this requirement might conflict with other dependencies' requirements. Or do we use any Django features that were deprecated/removed in 6.0.0?

PR #158, opened 2026-06-15
37: Upgrade psycopg to version 3 and enable database pooling

### LINKED ISSUE #37: Upgrade psycopg to 3.x
Django will deprecate `psycopg2`.

Upgrade to `psycopg` version 3.x. The problem currently is that `django-silk` has a bug that causes an error.

natkam:
Also: re-add the accidentally reverted commit where docker compose services and container names were updated.


aqueduct/aqueduct/settings.py
```diff
@@ -327,21 +328,29 @@ def my_transform(group: str, groups: list[str] | None = None) -> tuple[str, str]
...
+DATABASES: dict[str, Any] = {"default": {"ENGINE": DATABASE_ENGINE}}
 
 # 3. If it's SQLite, fill in the NAME for the file
 if DATABASE_ENGINE == "django.db.backends.sqlite3":
     DATABASES["default"]["NAME"] = str(BASE_DIR / "db.sqlite3")
     if TESTING:
-        DATABASES["default"]["TEST"] = {"NAME": str(BASE_DIR / "db_test.sqlite3")}  # type: ignore[assignment]
+        DATABASES["default"]["TEST"] = {"NAME": str(BASE_DIR / "db_test.sqlite3")}
 elif DATABASE_ENGINE == "django.db.backends.postgresql":
+    POSTGRESQL_POOL_MAX_SIZE = int(os.getenv("POSTGRESQL_POOL_MAX_SIZE", 20))
```
meffmadd, line 343: Specifying the `min_size` of the connection pool would be nice as well. We should add more config options rather than less in this PR rather than running into issues down the line.

meffmadd (changes requested):
More config options would be nice! `min_size` is the most important I think but others like `max_waiting` might be useful as well. If that is implemented it can be merged.

natkam:
I've added further connection pool options - the ones that seemed potentially relevant to me. The default values are the same as in the `psycopg_pool.ConnectionPool` class. Let me know if you think anything is missing!

Btw, I've also changed the required method for the token clean-up endpoint: it was `POST`, which was just copy-paste from the other endpoint, but didn't make much sense. I replaced it with `DELETE`.

PR #159, opened 2026-06-15
157: Log request times

### LINKED ISSUE #157: Save the total request time along with the client response time in `Request` model
Right now we cannot tell what overhead Aqueduct is producing.

Move the `@log_request` decorator to the innermost position possible for all endpoints. Pass `kwargs["request_start"]` in the first decorator. Then we set a `Request.total_response_time_ms` or similar based in the start time of the request and the end time.

This would also play nicely for #148

natkam:
To be merged after https://github.com/TU-Wien-dataLAB/aqueduct/pull/158 has been merged.

Until now, in `Request` objects, we've only measured the time of the view execution (and the external API query inside it). Now, the `response_time` also includes all the decorators, which seems more appropriate.

Additional, unrelated change: I've noticed that some type annotations in the `usage.py` file were wrong, so I've fixed them.


meffmadd (changes requested):
There seems to have been a misunderstanding about the purpose of this change... 

We should keep the `response_time_ms` calculation as is!
This should introduce an **additional** field on the Request model `request_time_ms` or `processing_time_ms` just as you implemented. This allows us to **calculate the overhead** spent in Aqueduct as `request_time_ms - response_time_ms` later in a future `/metrics` endpoint etc.

natkam:
@meffmadd I ended up naming the field `total_response_time_ms`, because I didn't want to rename the existing field `response_time_ms`, and it seemed to me like this more accurately expresses what the field contains. But I'm open to suggestions about both the field names and help texts!

In principle, it would also be possible to measure the time from the request processing start to the moment the view function is called in the `log_request` decorator - that would be the processing time, or the "Aqueduct overhead". But I guess it may also be useful to have the total time.

And: logging of the total time for streaming responses (in `_openai_stream`) is kinda clunky. I can try to think if there's a neater way to handle this, but for now, this solution works, so I've committed it as it is. Note: if we measure the processing/Aqueduct time instead of the total time, this adjustment won't be necessary at all.

Also: in the `usage` view, average response times for `completion` and `embedding` are calculated; I didn't add the respective average _total_ response times there, because it seems like adding any more stats would require changing the layout of that part of the page (that box is already quite wide). Let me know if you think it's worthwhile, or if that part is going to be rewritten anyway as part of the UI improvements. I can adjust the templates tomorrow.

meffmadd:
You are right in that case measuring the Aqueduct overhead + response time is cleaner. We should then take the end of the processing time as the exact start of the response time for a clean "hand-off". Also we do not need to track different processing times, just one for any request is enough.

meffmadd (approved):
LGTM

Issue #70, opened 2025-11-06
Track RPM and TPM through Django cache API


meffmadd:
**Is your feature request related to a problem? Please describe.**
Might be faster than db access. Queries on big request table are only made for usage analytics not for every request.

**Describe the solution you'd like**
Something like:
``` python
from django.core.cache import cache
from datetime import datetime

def track_request(identifier: str, limit_per_minute: int = 60) -> bool:
    """
    Tracks requests per minute per identifier (e.g. user_id, IP).
    Returns True if within limit, False if limit exceeded.
    """

    # Create a key unique to the minute and user
    current_minute = datetime.utcnow().strftime("%Y%m%d%H%M")  # e.g. 202511061430
    cache_key = f"rpm:{identifier}:{current_minute}"

    # Increment count atomically
    count = cache.incr(cache_key)
    
    # First time this key is created, set its expiration to 60 seconds
    if count == 1:
        cache.expire(cache_key, 60)

    # Return whether the request is still within limit
    return count <= limit_per_minute

```

PR #163, opened 2026-07-13
Cache Request Limits using Cache API with quotas shown in UI

meffmadd:
## Summary

Moves per-request rate limiting off SQL aggregates and onto the Django cache API, adds configurable **hourly/daily quota backstops**, and surfaces live quota usage in the UI (tokens page + admin). Implements #70 / #102.

## Background

`check_limits` previously ran 2 DB aggregate queries over `Request` (`Count`/`Sum`, plus a per-model count for the weighted request budget) on **every gateway request**. This PR replaces those queries with fixed **minute/hour/day cache buckets** keyed per token. The `Request` table now serves usage **analytics only** — the hot path no longer touches `management_request`.

## What changed

- **`gateway/rate_limiting.py` (new)** — cache-backed buckets (`{"req", "in", "out"}`) per `(token, window)`, TTL `2 × window`. The weighted request budget (`req`, scaled by per-model multipliers) is **reserved at check time**; input/output tokens are **recorded at completion** (non-streaming in `log_request`, streaming in `_openai_stream`). All writes run under a per-token advisory `cache_lock`; buckets are stored as pickled dicts, so the Redis float-`incr` limitation is avoided entirely.
- **Hourly/daily backstops** — new `hourly_limit_multiplier` / `daily_limit_multiplier` fields on `LimitMixin` (Org/Team/UserProfile), resolved through the existing hierarchy (specific → org → settings default `60`/`1440`). Hour/day limits are **derived** from the per-minute limits × multiplier, so a `None` per-minute cap implies no hour/day cap. `LimitMixin.clean()` bounds multipliers to `1–60` / `1–1440` and enforces `daily ≤ 24 × hourly`. Migration `0011_org_daily_limit_multiplier_and_more`.
- **Admin** — the two new fields are editable for Org/Team/UserProfile; the Token changelist gained **Hourly / Daily usage** columns (batch-fetched from the same buckets, no N+1).
- **Tokens page UI** — each token / service account now shows hourly and daily request-limit usage bars (`_rate_usage.html`), fetched in one `cache.get_many` per page (`get_per_token_usage`).
- **Settings** (defaults in braces) — `AQUEDUCT_RATE_LIMIT_ENABLED` (True), `AQUEDUCT_RATE_LIMIT_LOCK_TTL_SECONDS` (5), `AQUEDUCT_HOURLY_LIMIT_MULTIPLIER` (60), `AQUEDUCT_DAILY_LIMIT_MULTIPLIER` (1440). `CACHES` now falls back to `LocMemCache` under `DEBUG` too.
- **Ports #171** — 429 request-limit messages scale the limit by the model's multiplier (`Request limit (6/min)` style is preserved).

## Behavior changes to review

- **Sliding 60s → fixed minute/hour/day buckets** (accepted in #70): a burst straddling a window boundary gets fresh buckets.
- A request that passes the check but is rejected later (e.g. 400/404) now **reserves a request slot** — previously it did not (no `Request` row).
- Tokens are recorded into the **completion** window, so a request spanning a boundary counts its tokens in the window they were produced in.
- On rare same-token `cache_lock` contention the check **fails open** and token recording is **skipped** (both logged) — no worse than today's read races.

## Testing

- New `gateway/tests/test_rate_limiting.py` — bucket write/read, weighted budget across models, hour/day rollover (`60 reqs / 6 min → 61st blocked …/hour`), hierarchy + settings-default multiplier resolution, lock-contention fail-open/skip.
- Updated `TokenLimitTest` — `cache.clear()` isolation (DB deletion no longer resets counts); patched multiplier tests retargeted to `rate_limiting.get_model_request_limit_multiplier`.
- `management/tests` — `LimitMixin.clean()` bounds validation (`0`/`61`/`1441`, `daily > 24×hourly` rejected), tokens-page usage bars, admin hourly/daily columns.
- Full suite + `mypy` + pre-commit green.


meffmadd:
Closes #70

aqueduct/gateway/tests/test_endpoints.py
```diff
@@ -1591,6 +1592,35 @@ def test_per_model_expensive_multiplier_limits_requests(self, mock_multipliers):
...
 
+    def test_check_limits_uses_no_request_table_sql(self):
+        """A rate-limited (429) request issues no SQL against ``management_request``.
+
+        ``check_limits`` now reads from the cache, not the ``Request`` table. A
+        blocked request returns 429 before ``log_request`` runs, so it must not
+        touch ``management_request`` at all (no aggregate SELECT, no INSERT).
+        """
+        from django.db import connection
+        from django.test.utils import CaptureQueriesContext
```
natkam: please move the imports to the top of the file

aqueduct/gateway/rate_limiting.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+def _cache_lock(lock_id: str, ttl: int) -> AbstractContextManager[bool]:
+    """Lazy import of ``cache_lock`` to avoid a circular import at module load.
+
+    ``gateway.views.utils`` lives in the ``gateway.views`` package whose ``__init__``
+    eagerly imports the view modules (which, in turn, import this module). Importing
+    it at module top-level would therefore create a load-order cycle. By the time
+    these functions run (request time) the app is fully loaded, so the import is a
+    cheap dict lookup.
```
natkam: I don't think we need explaining how circular imports work here. The first line of the docstring really says all a developer has to know about this function, and is much faster to skim than this description, which is longer than the function's body.

aqueduct/gateway/rate_limiting.py
```diff
@@ -0,0 +1,253 @@
...
+                break  # finest exceeded window for this metric; skip coarser windows
+    return messages
+
+
+def check_and_reserve(limits: LimitSet, token_id: int, model: str | None) -> tuple[bool, list[str]]:
+    """Read all-window buckets, enforce limits, reserve the request slot in every window.
+
+    Returns ``(allowed, exceeded_messages)``. On rare same-token lock contention we
+    fail open (allow the request) to avoid latency spikes — no worse than today's
+    read races.
```
natkam: What are "today's read races" supposed to mean tomorrow? or when this code is merged?

aqueduct/gateway/rate_limiting.py
```diff
@@ -0,0 +1,253 @@
...
+    # The model's request-limit multiplier drives both the reservation delta
+    # (each request consumes 1/multiplier of budget) and the request limit shown in
+    # a 429 message (base limit x multiplier = effective capacity for this model).
+    request_multiplier = get_model_request_limit_multiplier(model) if model else 1.0
+    delta = 1.0 / request_multiplier
+
+    now = timezone.now()
+    keys = [_bucket_key(token_id, name, now) for name, _secs, *_rest in windows]
+
+    with _cache_lock(_lock_key(token_id), settings.AQUEDUCT_RATE_LIMIT_LOCK_TTL_SECONDS) as got:
```
natkam: Why `got`? Is it some abbreviation?

Or, if it's just past participle of "get", maybe "aquired" would be a more fortunate name?

aqueduct/gateway/views/utils.py
```diff
@@ -80,6 +81,10 @@ async def _stream() -> AsyncGenerator[str, None]:
         request_log.token_usage = token_usage
         request_log.response_time_ms = int((end_time - start_time) * 1000)
         await request_log.asave()
+        # Record token usage into the rate-limit buckets. Streaming requests
+        # defer recording to here (stream end) since token usage is only known
+        # once the upstream stream completes.
+        record_token_usage(request_log.token_id, request_log.token_usage)
```
natkam: Shouldn't this method be wrapped in `sync_to_async`, just like in the decorators?

aqueduct/management/views/tokens.py
```diff
@@ -39,11 +40,21 @@ def get_context_data(self, **kwargs) -> dict[str, object]:
         # Effective limits for user-owned tokens (UserProfile + Org fallback)
         user_limits: LimitSet = LimitSet.from_objects(profile, profile.org)
 
+        # Current rate-limit usage per token (shown under each list item). Each
+        # token operates under ``user_limits`` independently; the bars report
+        # that token's own consumed request budget. Fetch every user-token
+        # bucket in a single ``cache.get_many``.
```
natkam: This comment references the page layout, which seems out of place here. (Also, the wording is bizarre.) How about:
```
# Attach each token's current rate-limit usage (per minute/hour/day bucket)
# to the token object so templates can render it. All tokens are fetched in a
# single ``cache.get_many`` call to avoid one cache lookup per token.
```

aqueduct/management/admin.py
```diff
@@ -597,10 +610,79 @@ def team_link(self, obj) -> str:
...
+        """
+        if token.service_account:
+            team = token.service_account.team
+            return LimitSet.from_objects(team, team.org)
+        profile = token.user.profile
+        return LimitSet.from_objects(profile, profile.org)
+
+    def _usage_percent(self, token: Token, window_name: str, window_index: int) -> str:
+        limits = self._token_limits(token)
+        # windows()[window_index] = (name, secs, rpm, itpm, otpm); [2] = requests-per-window.
```
natkam: Is it `requests-per-window`, or `requests-per-minute`?

aqueduct/management/tests/test_limit_validation.py
```diff
@@ -0,0 +1,147 @@
+"""Tests for the ``LimitMixin`` hourly/daily multiplier fields and validation.
+
+Covers §3.3/§4.2 of the cache-RPM plan:
```
natkam: What cache-RPM plan is meant here? Is it included in the repo? Or can it be linked? If I can't access it, its mention is not very helpful, and can actually be plain wrong if the plan is updated and the paragraph numbers change.

aqueduct/management/tests/test_limit_validation.py
```diff
@@ -0,0 +1,147 @@
...
+        LITELLM_ROUTER_CONFIG_FILE_PATH=Path(__file__).resolve().parents[3]
+        / "example_router_config.yaml"
+    )
+    def test_full_clean_surfaces_multiplier_error_via_admin_form(self):
+        """The admin ModelForm runs full_clean(), surfacing the clean() bounds.
+
+        The form supplies the JSON exclusion fields, so full_clean() reaches
+        clean() (a bare model.full_clean() would trip on the blank default []).
+        """
+        from management.admin import OrgAdminForm
```
natkam: https://docs.astral.sh/ruff/rules/module-import-not-at-top-of-file/

aqueduct/management/tests/test_limit_validation.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+        self.assertFalse(form.is_valid())
+        self.assertIn("hourly_limit_multiplier", form.errors)
+
+
+class AdminFormFieldsTest(TestCase):
+    """The new fields are editable in the Django admin forms (lightweight check)."""
+
+    def test_fields_present_in_admin_forms(self):
+        from management.admin import OrgAdminForm, TeamAdminForm, UserProfileAdminForm
```
natkam: https://docs.astral.sh/ruff/rules/module-import-not-at-top-of-file/

aqueduct/gateway/rate_limiting.py
```diff
@@ -0,0 +1,253 @@
...
+    Returns ``(allowed, exceeded_messages)``. On rare same-token lock contention we
+    fail open (allow the request) to avoid latency spikes — no worse than today's
+    read races.
+    """
+
+    windows = limits.windows()
+    if not any(
+        rpm is not None or itpm is not None or otpm is not None
+        for _name, _secs, rpm, itpm, otpm in windows
+    ):
```
natkam: Actually, there already is a function that does exactly this: `_has_any_limit` in `decorators.py`.

aqueduct/management/tests/test_tokens_page.py
```diff
@@ -0,0 +1,95 @@
+"""Tests for the tokens list page — per-window request-usage progress bars.
+
+Covers the ``_rate_usage.html`` include driven by
+``gateway.rate_limiting.get_per_token_usage`` on ``UserTokensView``, rendered
+under each API key list item as two daisyUI linear ``progress`` bars
```
natkam: While testing the values passed to the template and displayed in the html makes a lot of sense, testing that "two daisyUI linear ``progress`` bars" are rendered seems a tad too specific. Isn't it just a cosmetic detail? Or is it crucial for the page to function that we use daisyUI, and that the bars are styled with the "progress-primary" CSS class (cf. assertions in the tests)?

natkam (changes requested):
Mostly complaints about AI-generated comments and docstrings, which are a PITA to read, so sorry if they come across as somewhat brusque. Nothing personal here :'D

The one with `sync_to_async` may be something more substantial, if we use Redis for caching (I guess it doesn't make any difference with the in-memory cache?) - but have a look at it yourself.

Oh, and speaking of AI-generated gibberish: the PR description mentions "the hot path no longer touches `management_request`", which - depending on what is actually meant by "the hot path" - may be wrong or misleading, because processing of a request still saves the `Request` object to the database at least once.

Which leads me to a (genuine) question, what the point of an AI-generated PR description is. If I, as a reviewer, need one, I can ask an LLM to generate it for me; it's much less time and effort than actually reading through one. This description is unnecessarily verbose, uses weird words and expressions, and, as I said, is a pain to read. I would find a shorter, but human-written (or at least human-polished) description much more helpful.

If that description is meant for agents and not humans, then just ignore the previous paragraph. :grin: Otherwise, I wouldn't mind discussing this topic.

natkam (approved):

PR #171, opened 2026-07-18
Show scaled request limit in 429 errors

### LINKED ISSUE #167: Request limit 429 errors show the base request limit instead of values based on multiplier
**Describe the bug**
When a user runs into a request limit with a model that has a request limit multiplier, the error message shows the base request limit instead of the scaled value for the specific model. This is confusing since the user would expect the limit to match with their experience.

**To Reproduce**
1. Set request limit multiplier on model (low for easier testing)
2. Trigger the limit by sending requests
3. Observe 429 response text mention base limit

**Expected behavior**
Should be scaled to the model the request tried to use.

floze-the-genius:
Closes #167

## Summary

- apply the current model request-limit multiplier to the limit shown in 429 responses
- resolve model aliases before reading the multiplier
- preserve compact output for both whole and fractional scaled limits
- add regression assertions for 2x and 0.5x model multipliers

## Root cause

Rate-limit enforcement already used a weighted request count, but the error message always interpolated the unscaled base `requests_per_minute` value. This made the reported quota disagree with the quota users experienced for models with a multiplier.

## Validation

- `uv run python manage.py test gateway.tests.test_endpoints.TokenLimitTest` (11 passed)
- focused multiplier tests (2 passed)
- `uv run ruff check`
- `uv run ruff format --check`
- `uv run mypy`
- `uv run pre-commit run --all-files`

The full `gateway` suite ran 194 tests locally; three unrelated MCP live-server classes failed during macOS multiprocessing startup with `AppRegistryNotReady` / `Server stopped`. The focused suite and all static checks pass, and upstream Linux CI covers the full gateway lane.

## Coordination

Draft PR #163 refactors rate-limit storage and currently retains the same base-limit message behavior. This patch targets current `main`; if #163 lands first, the message calculation will need a mechanical rebase into its new rate-limiting module.

## AI assistance disclosure

This contribution was implemented, tested, and documented with OpenAI Codex assistance under the submitter's direction.

### DIFF
--- a/aqueduct/gateway/tests/test_endpoints.py
+++ b/aqueduct/gateway/tests/test_endpoints.py
@@ -1553,6 +1553,9 @@ class TokenLimitTest(ChatCompletionsBase):
             429,
             f"Expected 429 Too Many Requests, got {response7.status_code}: {response7.content}",
         )
+        self.assertEqual(
+            response7.json()["error"]["message"], "Rate limit exceeded. Request limit (6/min)."
+        )
 
     @patch(
         "gateway.views.decorators.get_all_model_request_limit_multipliers",
@@ -1584,6 +1587,9 @@ class TokenLimitTest(ChatCompletionsBase):
         # Third request: weighted = 4.0 >= 3
         response3 = self._send_chat_completion(self.MESSAGES, max_completion_tokens=5)
         self.assertEqual(response3.status_code, 429)
+        self.assertEqual(
+            response3.json()["error"]["message"], "Rate limit exceeded. Request limit (1.5/min)."
+        )
 
 
 @override_settings(LITELLM_ROUTER_CONFIG_FILE_PATH="router.yaml")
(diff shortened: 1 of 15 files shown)


meffmadd (approved):
LGTM

meffmadd:
@floze-the-genius Thanks for the PR! Could you fix the linting errors then I will merge.

floze-the-genius:
Rebased onto current main and fixed the requested Ruff LOG004 findings by using `log.error` outside exception handlers. Validation: `uv run ruff check .` and `uv run python manage.py test gateway.tests.test_endpoints.TokenLimitTest` (11 passed). Ready for merge.

meffmadd:
Still fails unfortunately... Could you run the `ruff` formatter locally without the `check` flag to fix those last errors?

floze-the-genius:
Fixed in 9db2484. The root cause was version skew: the project lock runs Ruff 0.15.5 and reported clean, while CI resolved the >=0.15.5 requirement to Ruff 0.16.0, which reformats Python snippets in 12 Markdown files. I applied Ruff 0.16.0 formatting and re-ran the exact CI checks: ruff check, import sorting, and format check all pass. The targeted TokenLimitTest suite also passes (11 tests). CI is rerunning now.

PR #173, opened 2026-07-27
Linting issues; missing `@tos_accepted` decorator; update GH pipeline actions

natkam:
- In some of the views, `@tos_accepted` decorator was missing. It seems to me that it must have been an oversight; or do you know if there was a reason why it wasn't added?
- I've fixed deprecation warnings from the GH actions: multiple actions depended on an old version of Node, so I just updated them all to the newest version.
- The attestation action complained about a permission to write artifact metadata it needed; the docs also stated that this permission should be enabled, so I added it to the list. 
- A few days ago a new version of ruff was released (0.16); it now by default formats code snippets in Markdown files - so these are just formatting changes. It also (rightly) complained about logging `log.exception()` where no exception had been raised.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -20,22 +20,22 @@ jobs:
       artifact-metadata: write
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
       # Uses the `docker/login-action` action to log in to the Container registry
       # using the account and password that will publish the packages. Once published,
       # the packages are scoped to the account defined here.
       - name: Log in to the Container registry
-        uses: docker/login-action@v4
+        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f  # v4.6.0
         with:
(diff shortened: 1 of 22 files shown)


.github/workflows/docker.yml
```diff
@@ -17,35 +17,45 @@ jobs:
...
       - name: Checkout repository
-        uses: actions/checkout@v4
-      # Uses the `docker/login-action` action to log in to the Container registry registry using the account and password that will publish the packages. Once published, the packages are scoped to the account defined here.
+        uses: actions/checkout@v7
+      # Uses the `docker/login-action` action to log in to the Container registry
+      # using the account and password that will publish the packages. Once published,
+      # the packages are scoped to the account defined here.
       - name: Log in to the Container registry
-        uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
+        uses: docker/login-action@v4
```
meffmadd: We should still use hashes to avoid supply chain attacks.

meffmadd:
Otherwise it can be merged.

.github/workflows/docker.yml
natkam: Good point, I've updated the tags.

PR #178, opened 2026-08-10
145: Pass raw claims dict to get_org_name_from_groups (Part 1/3)

JaeYeonLee0621:
1. Changed the `groups` parameter to `claims` in organization mapping resolution logic.
2. Renamed `get_org_name_from_groups` to `get_org_name` to reflect the updated claims signature.
3. Updated docstrings and comments.
4. Updated existing unit tests and added test cases.

### DIFF
--- a/aqueduct/management/auth.py
+++ b/aqueduct/management/auth.py
@@ -22,21 +22,21 @@ def default_org_name_from_groups(groups: list[str]) -> str | None:
     return groups[0]
 
 
-def get_org_name_from_groups(groups) -> str | None:
+def get_org_name_from_groups(claims) -> str | None:
     """
     Extracts the organization name from the user's groups.
     """
     if hasattr(settings, "ORG_NAME_FROM_OIDC_GROUPS_FUNCTION"):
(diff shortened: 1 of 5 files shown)


aqueduct/management/auth.py
```diff
@@ -12,32 +12,31 @@
...
-    Default implementation to extract organization name, which returns the first group in the list.
-    Override this or set ORG_NAME_FROM_OIDC_GROUPS_FUNCTION in settings.
+    Default implementation to extract organization name from claims.
+    Returns the first group in the claims['groups'] list.
     """
-    if not groups:
-        return None
-    return groups[0]
+    if groups := claims.get("groups"):
+        return groups[0]
```
meffmadd: The logic here changed a bit. The function is not really defensive anymore. `groups` can be anything so `groups[0]` can fail arbitrarily.

I think this function was flawed in general. I think it should just return "default" as the default group so the return type just is `Literal["default"]`...

meffmadd (changes requested):

meffmadd (approved):
LGTM!

PR #179, opened 2026-08-10
145: Pass raw claims dict to _get_teams_from_groups (Part 2/2)

### LINKED ISSUE #145: Pass all OAuth claims to auth mapping functions
We use configurable functions to get teams, orgs etc. from the OAuth info. For now these functions only accept `groups`, which might not be in the claims by an IdP.

The functions should instead accept arbitrary claims and the admins can then configure the functions as they need them. In the worst case this will be `claims.get("groups")`.

JaeYeonLee0621:
1. Changed the groups parameter to claims in organization mapping resolution logic.
2. Renamed `_get_teams_from_groups` to `_get_teams` to reflect the updated claims signature.
3. Updated docstrings and comments.
4. Updated existing unit tests and added test cases.

### DIFF
--- a/aqueduct/aqueduct/settings.py
+++ b/aqueduct/aqueduct/settings.py
@@ -137,7 +137,7 @@ ENABLE_OAUTH_GROUP_CREATION = os.getenv("ENABLE_OAUTH_GROUP_CREATION", "True").l
 ENABLE_OAUTH_GROUP_REMOVAL = os.getenv("ENABLE_OAUTH_GROUP_REMOVAL", "True").lower() == "true"
 
 
-def default_oauth_team_names_from_groups(
+def default_oauth_team_names(
     group: str, groups: list[str] | None = None
 ) -> tuple[str, str] | None:
     """
@@ -160,7 +160,7 @@ def default_oauth_team_names_from_groups(
     return None
 
 
-OAUTH_TEAM_NAMES_FROM_GROUPS_FUNCTION = default_oauth_team_names_from_groups
+OAUTH_TEAM_NAMES_FUNCTION = default_oauth_team_names
 
 EXTRA_NAV_LINKS = {
     "Bug Report": "https://github.com/TU-Wien-dataLAB/aqueduct/issues/new?template=bug_report.md",
--- a/aqueduct/aqueduct/settings.py
+++ b/aqueduct/aqueduct/settings.py
@@ -138,30 +138,47 @@ ENABLE_OAUTH_GROUP_CREATION = os.getenv("ENABLE_OAUTH_GROUP_CREATION", "True").l
 ENABLE_OAUTH_GROUP_REMOVAL = os.getenv("ENABLE_OAUTH_GROUP_REMOVAL", "True").lower() == "true"
 
 
-def default_oauth_team_names(claims) -> list[tuple[str, str]]:
+def default_oauth_team_names(claims) -> list[str]:
     """
-    Default function to transform OAuth claims to team mappings.
+    Default function to extract team names from OAuth claims.
 
     Args:
-        claims: The full OAuth claims dict containing 'groups' and other user info
+        claims: The full OAuth claims dict.
 
     Returns:
-        List of (team_name, original_group_name) tuples, or empty list if none.
+        List of team names, or empty list if none.
 
     Example:
-        def my_transform(claims) -> list[tuple[str, str]]:
-            groups = claims.get('groups', [])
+        def my_extract_team_names(claims) -> list[str]:
+            team_names = claims.get('groups', [])
+            return [team_name for team_name in team_names if team_name.startswith('E')]
+    """
+    return []
+
+
+def default_oauth_display_team_names(team_names: list[str]) -> list[tuple[str, str]]:
+    """
+    Default function to map team names to display team names.
+
+    Args:
+        team_names: List of team names.
+
+    Returns:
+        List of (display_team_name, team_name) tuples, or empty list if none.
+
+    Example:
+        def my_display_team_names(team_names) -> list[tuple[str, str]]:
             result = []
-            for group in groups:
-                if group.startswith('E'):
-                    team_name = group.split('-')[0]
-                    result.append((team_name, group))
+            for team_name in team_names:
+                display_team_name = team_name.split('-')[0]
+                result.append((display_team_name, team_name))
             return result
     """
     return []
 
 
 OAUTH_TEAM_NAMES_FUNCTION = default_oauth_team_names
+OAUTH_DISPLAY_TEAM_NAMES_FUNCTION = default_oauth_display_team_names
 
 EXTRA_NAV_LINKS = {
     "Bug Report": "https://github.com/TU-Wien-dataLAB/aqueduct/issues/new?template=bug_report.md",
--- a/aqueduct/aqueduct/settings.py
+++ b/aqueduct/aqueduct/settings.py
@@ -13,7 +13,6 @@ https://docs.djangoproject.com/en/5.2/ref/settings/
 import logging
 import os
 import sys
-from datetime import datetime
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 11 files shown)


aqueduct/aqueduct/settings.py
```diff
@@ -137,7 +137,7 @@ def my_org_name_extractor(groups: list[str]) -> str | None:
 ENABLE_OAUTH_GROUP_REMOVAL = os.getenv("ENABLE_OAUTH_GROUP_REMOVAL", "True").lower() == "true"
 
 
-def default_oauth_team_names_from_groups(
+def default_oauth_team_names(
     group: str, groups: list[str] | None = None
```
meffmadd: Not sure what the plan is for PR 3/3 but this should take `claims: dict[Any,Any]` (or `claims: dict[str, Any]` depending on the OIDC spec allows this).

aqueduct/management/tests/test_oauth_team_creation.py
```diff
@@ -12,9 +12,7 @@
 User = get_user_model()
 
 
-def sample_team_names_from_groups(
-    group: str, groups: list[str] | None = None
-) -> tuple[str, str] | None:
+def sample_team_names(group: str, groups: list[str] | None = None) -> tuple[str, str] | None:
```
meffmadd: Same here right? We should also pass claims here eventually.

aqueduct/management/auth.py
```diff
@@ -54,10 +54,10 @@ def _get_teams_from_groups(self, groups: list[str]) -> list[tuple[str, str]]:
         if not getattr(settings, "ENABLE_OAUTH_GROUP_MANAGEMENT", False):
             return []
 
-        func = getattr(
-            settings, "OAUTH_TEAM_NAMES_FROM_GROUPS_FUNCTION", lambda group, groups=None: None
-        )
+        func = getattr(settings, "OAUTH_TEAM_NAMES_FUNCTION", lambda group, groups=None: None)
+
         team_mappings = []
+        groups = claims.get("groups", [])
```
meffmadd: This is exactly what we want to avoid! We should get the team names from the `claims` directly as well. We do not want to assume that groups (from which we get the team names) are called `groups` in the claim.

meffmadd (changes requested):

aqueduct/management/admin.py
```diff
@@ -33,6 +34,15 @@
 log = logging.getLogger("aqueduct")
```
JaeYeonLee0621, line 36: Ruff reported that `sync_oauth_team_names_action` is too complex, so I divided the logic into two internal methods.

1. Fetch the new team-to-group mapping.
2. Prepare the sync data.

Since the second method returns three lists, I created a dataclass for the return type rather than using a tuple of lists.

If you think this approach is still overcomplicating things, I’m open to simpler ways to refactor it. Let me know what you think :)

aqueduct/management/admin.py
```diff
@@ -178,104 +188,121 @@ def delete_tos_cache(modeladmin, request, queryset):
         cache.delete(f"django:tos:agreed:{user.id}", version=key_version)
 
 
+def _get_new_mapping() -> dict[str, str]:
+    if not (func := getattr(settings, "OAUTH_TEAM_NAMES_FUNCTION", None)):
+        raise OAuthConfigurationError("OAUTH_TEAM_NAMES_FUNCTION not configured")
```
JaeYeonLee0621, line 193: I created two separate error types to make the error messages more precise. However, if this feels overcomplicated, please let me know!

aqueduct/management/admin.py
```diff
@@ -178,104 +188,121 @@ def delete_tos_cache(modeladmin, request, queryset):
...
+
+    all_oauth_groups = list(
+        Team.objects.exclude(oauth_group_name="")
+        .values_list("oauth_group_name", flat=True)
+        .distinct()
+    )
+    claims = {"groups": all_oauth_groups}
```
JaeYeonLee0621, line 200: I’m stuck on how to handle this part. I need claims to pass into the function, but there’s currently no way to access them here.

Assuming claims has a groups key feels out of scope for this issue. Alternatively, we could use custom OAUTH_GROUP_FUNCTION from settings, but I'm not sure if that's a good idea.

I’d appreciate any thoughts or suggestions on this.

aqueduct/management/admin.py
meffmadd: We do not do this really so far and there would be a lot of places where custom error types would make sense. I think we should move this to a separate PR one day.

aqueduct/management/admin.py
meffmadd: Ok that's a case I did not think about... What if we split the single `OAUTH_TEAM_NAMES_FUNCTION` into two parts/functions? One function handles converting `claims` to a list of "raw" team names and one function that maps the "raw" team names to team "display names". Then in the admin UI we only use the second function for updates.

aqueduct/management/admin.py
meffmadd: Yeah I really like that!

meffmadd (changes requested):
We still have to crack the case that you raised with the admin updates. Otherwise, the changes are very clean!

aqueduct/aqueduct/settings.py
```diff
@@ -132,35 +131,54 @@ def my_org_name_extractor(groups: list[str]) -> str | None:
 # OAuth Group Management Settings
 # Controls automatic team creation and membership management from OAuth groups
 
-ENABLE_OAUTH_GROUP_MANAGEMENT = os.getenv("ENABLE_OAUTH_GROUP_MANAGEMENT", "False").lower() == "true"
+ENABLE_OAUTH_GROUP_MANAGEMENT = (
+    os.getenv("ENABLE_OAUTH_GROUP_MANAGEMENT", "False").lower() == "true"
+)
 ENABLE_OAUTH_GROUP_CREATION = os.getenv("ENABLE_OAUTH_GROUP_CREATION", "True").lower() == "true"
 ENABLE_OAUTH_GROUP_REMOVAL = os.getenv("ENABLE_OAUTH_GROUP_REMOVAL", "True").lower() == "true"
```
JaeYeonLee0621, line 139: I split the function into two steps (`extracting team names` and `converting them to display names`).

However, the naming is currently a bit inconsistent, in some parts of the code, `original_group_name` is used same meaning as `team_name`.

If we want to standardize the terminology, I can rename the variables and functions from `team_name` to `original_group_name` or `group_name`.

charts/aqueduct/files/settings.py
```diff
@@ -132,35 +131,37 @@ def my_org_name_extractor(groups: list[str]) -> str | None:
...
+    os.getenv("ENABLE_OAUTH_GROUP_MANAGEMENT", "False").lower() == "true"
+)
 ENABLE_OAUTH_GROUP_CREATION = os.getenv("ENABLE_OAUTH_GROUP_CREATION", "True").lower() == "true"
 ENABLE_OAUTH_GROUP_REMOVAL = os.getenv("ENABLE_OAUTH_GROUP_REMOVAL", "True").lower() == "true"
 
 
-def default_oauth_team_names_from_groups(
-    group: str, groups: list[str] | None = None
-) -> tuple[str, str] | None:
+def default_oauth_team_names(claims) -> list[tuple[str, str]]:
```
JaeYeonLee0621: There is another function here that needs to be split.

If we want to keep the logic identical between the two settings file, 
I think it would be better to extract it into a separate utility file and import it into the settings.py files. 🤔 

Let me know your thoughts on this!

aqueduct/aqueduct/settings.py
meffmadd: Naming is fine for now. Maybe `OAUTH_TEAM_DISPLAY_NAMES_FUNCTION` but both are fine either way.

aqueduct/management/auth.py
```diff
@@ -39,48 +39,62 @@ def _org(self, claims) -> Org | None:
...
+        map_func = getattr(settings, "OAUTH_DISPLAY_TEAM_NAMES_FUNCTION", lambda names: None)
+
+        try:
+            result = map_func(team_names)
+        except Exception as e:
+            log.exception("Error calling OAUTH_DISPLAY_TEAM_NAMES_FUNCTION: %s", e)
+            return []
+
+        if result is None:
+            return []
```
meffmadd: This is not necessary since the next `if` catches `None` as well.

charts/aqueduct/files/settings.py
meffmadd: The settings file for the chart is kept in sync with the "main" settings file using a pre-commit hook automatically!

meffmadd (approved):
Small nitpicks you can still fix but overall: LGTM!

aqueduct/aqueduct/settings.py
meffmadd: Oh now I know what you mean. Yes this should be called `ENABLE_OAUTH_GROUP_CREATION` and `ENABLE_OAUTH_GROUP_REMOVAL`.

meffmadd (changes requested):

meffmadd (approved):
LGTM!

aqueduct/aqueduct/settings.py
```diff
@@ -132,35 +131,54 @@ def my_org_name_extractor(groups: list[str]) -> str | None:
...
+
+    Example:
+        def my_display_team_names(team_names) -> list[tuple[str, str]]:
+            result = []
+            for team_name in team_names:
+                display_team_name = team_name.split('-')[0]
+                result.append((display_team_name, team_name))
+            return result
+    """
+    return []
```
meffmadd: I would be nice if the default function would return the input team names unmodified as output team names as well. That would be a good default so that admins if needed only have to define OAUTH_TEAM_NAMES_FUNCTION if that is enough.

```suggestion
    return [(t, t) for t in team_names]
```

meffmadd (changes requested):

PR #180, opened 2026-08-10
145: Remove _groups method from OIDCBackend (Part 3/3)

### LINKED ISSUE #145: Pass all OAuth claims to auth mapping functions

JaeYeonLee0621:


aqueduct/management/auth.py
```diff
@@ -199,7 +196,7 @@ def _sync_teams(self, user: User, profile: UserProfile, groups: list[str]) -> No
                     )
 
     def create_user(self, claims) -> User | None:
-        groups = self._groups(claims)
+        groups = claims.get("groups", settings.OIDC_DEFAULT_GROUPS)
```
JaeYeonLee0621, line 199: Delete it after #178 & #179 are merged
- It is used for _org def's parameter
- But #178 & #179 will remove groups parameter to claims

aqueduct/management/auth.py
JaeYeonLee0621, line 209: Extracting groups can be customized by each pre-defined function in settings
But in here, `groups` variable is used to figure out whether group is admin
If `claims.get("group")` is implemented here, extracting groups logic can be fragmented

aqueduct/management/auth.py
```diff
@@ -231,7 +228,7 @@ def create_user(self, claims) -> User | None:
 
     def update_user(self, user, claims) -> User:
         """Update existing user with new claims, if necessary save, and return user"""
-        groups = self._groups(claims)
+        groups = claims.get("groups", settings.OIDC_DEFAULT_GROUPS)
```
JaeYeonLee0621, line 231: Delete it after #178 & #179 are merged
- It is used for _org def's parameter
- But #178 & #179 will remove groups parameter to claims

aqueduct/management/auth.py
meffmadd: Ok so what is missing from the sets of PRs is the same mechanism that we now have for "Teams", we need for "Orgs" as well. Basically we will stop dealing in "groups" anymore but we just define functions that give us org name and team names.

meffmadd (changes requested):

PR #183, opened 2026-08-24
143: Replace OIDC-mapping settings functions with admin config snippets

### LINKED ISSUE #143: Make auth functions configurable via Snippets in DB + Plugin setup

Several config options in `settings.py` require the admin to configure callable functions (e.g. `ORG_NAME_FROM_OIDC_FUNCTION`, `OAUTH_TEAM_NAMES_FUNCTION`, `OAUTH_DISPLAY_TEAM_NAMES_FUNCTION`). Today that means editing a deployed `settings.py`, which cannot be changed at runtime.

Store configuration snippets in the **database as a model**, editable at runtime by superusers in the Django admin (like pipes in OpenWebUI).

⚠️ A superuser-editable Python snippet is arbitrary server-side code execution for superusers. Accepted for our self-hosted deployment; no sandboxing is planned.

## Decided design

- **Snippets are Python classes, not functions** — sets up the plugin infrastructure in #130 (which this issue blocks). Two types on one model:
  - `config` — the OIDC auth hooks (org / teams / user group)
  - `plugin` — pre/post processing (model + validation + admin surface now; **execution deferred to #130**)
- A **default base class** implements every config method with today's behavior; admins **subclass it and override** only what they need. When no active DB config exists, the default class is used (never fail-closed).
- **Config class methods** (`claims`-based):
  - `org_name(claims)` — org extraction (fail closed ⇒ no org, no login)
  - `team_names(claims)` — team names (fail open ⇒ log + `[]`)
  - `display_team_names(team_names)` — display mapping (fail open ⇒ log + `[]`)
  - `user_group(claims)` — sets `UserProfile.group`; enables org-admins via claims
- **Model:** `Snippet(name, type[config|plugin], active, code)`. Multiple snippets allowed; **at most one active `config`**; plugins identified by `name` + `active`.
- **New module `management/snippets.py`** used by both `auth.py` and the admin; the base class is auto-injected into the snippet's exec namespace.
- Resolution cached per process (`(name, updated_at)`); ~1 DB read per login.
- Tests move from `override_settings(...FUNCTION=...)` to the DB path.

## Admin

- Django admin, superuser-only; validation at save (compile + subclass/method-signature check).
- Test-drive console: paste an input string → see the method's output.

Open implementation details (user_group→flags mapping, uniqueness of one-active-config, plugin priority) live in `DESIGN-STATE-ISSUE-143.md` + `docs/adr/0001-snippets-in-database.md`.

JaeYeonLee0621:
- Added a new `ADMIN_SUPERUSER_EMAILS` setting                                               
- `SnippetAdmin` is restricted to superusers whose email is on `ADMIN_SUPERUSER_EMAILS`                                                                                                                                                                                                                            
- Create base `ConfigSnippet` class in management/snippets.py


aqueduct/aqueduct/settings.py
```diff
@@ -137,48 +135,10 @@ def my_org_name_extractor(groups: list[str]) -> str | None:
...
-    """
-    return [(t, t) for t in team_names]
-
-
-OAUTH_TEAM_NAMES_FUNCTION = default_oauth_team_names
-OAUTH_DISPLAY_TEAM_NAMES_FUNCTION = default_oauth_display_team_names
+ADMIN_GROUP = os.getenv("ADMIN_GROUP", "")
+ADMIN_SUPERUSER_EMAILS = [
+    e.strip().lower() for e in os.getenv("ADMIN_SUPERUSER_EMAILS", "").split(",") if e.strip()
+]
```
JaeYeonLee0621, line 141: 📝 

1. `ADMIN_GROUP` : is_staff + is_superuser
2. `ADMIN_SUPERUSER_EMAILS` : is_staff + is_superuser + access Snippet Admin

aqueduct/management/templates/admin/management/snippet/test_console.html
```diff
@@ -0,0 +1,108 @@
+{% extends "admin/base_site.html" %}
```
JaeYeonLee0621: For reference, this test console looks like below image

Image

aqueduct/management/snippets.py
```diff
@@ -0,0 +1,99 @@
...
+        if not callable(method):
+            raise ValidationError(f"Snippet class must define a '{method_name}' method.")
+        params = list(inspect.signature(method).parameters.values())
+        if len(params) != METHOD_SELF_PLUS_ARG_COUNT or params[0].name != "self":
+            raise ValidationError(
+                f"'{method_name}' must be an instance method taking exactly one "
+                f"argument, e.g. 'def {method_name}(self, claims)'."
+            )
+
+
```
JaeYeonLee0621, line 92: 📝 Skipped caching implementation
1. Direct read requests are already cheap and fast.
2. Latency benefits are minimal compared to the risk of stale cache states.

aqueduct/management/templates/admin/management/snippet/test_console.html
```diff
@@ -0,0 +1,108 @@
...
+    }
+  </style>
+{% endblock %}
+
+{% block content %}
+<div id="content-main">
+  <div class="module">
+    <p style="color:#b91c1c;font-weight:600;margin-bottom:1rem;">
+      <strong>Warning:</strong> {{ security_warning }} For this reason this console
+      (and snippet editing) is restricted to superusers.
```
natkam: Wouldn't it make more sense to store the whole warning text in the `security_warning`? Otherwise it may turn out that the whole text displayed on the page is incoherent.

Or actually, why pass it in as a template variable? It doesn't look like we use it anywhere else, do we?

aqueduct/management/tests/test_snippet_admin.py
```diff
@@ -0,0 +1,276 @@
...
+
+    @override_settings(ADMIN_SUPERUSER_EMAILS=["someone-else@example.com"])
+    def test_admin_group_superuser_not_on_allowlist_forbidden(self):
+        # Simulate an account that is promoted to admin group (thus superuser)
+        # via admin-group login but is NOT on ADMIN_SUPERUSER_EMAILS.
+        # Being a superuser is necessary but NOT sufficient for the snippet admin.
+        user = User.objects.create_superuser(username="you", email="you@example.com", password="pw")
+        self.client.force_login(user)
+        resp = self.client.get(self.changelist_url)
+        self.assertEqual(resp.status_code, 403)
```
natkam, line 257: I may be missing something, but how is this test different from the previous one (`test_superuser_not_in_allowlist_forbidden_on_changelist`)? Also, I'm not sure if I understand what the comment refers to. Can you have a look and confirm that all is right here?

aqueduct/management/tests/test_snippet_admin.py
```diff
@@ -0,0 +1,276 @@
...
+            username="admin", email=SUPERUSER_EMAIL, password="pw"
+        )
+        self.client.force_login(self.superuser)
+        self.url = reverse("admin:management_snippet_test_console")
+
+    def test_get_shows_console(self):
+        resp = self.client.get(self.url)
+        self.assertEqual(resp.status_code, 200)
+        self.assertContains(resp, "Snippet test console")
+        self.assertContains(resp, "executes real server code")
```
natkam, line 273: OK, now, this could be an argument to keep the variable with the warning text, `SNIPPET_CONSOLE_SECURITY_WARNING`. If we have it, we can use it to make assertions in this test, instead of using the hard-coded parts of its value.

docs/user-guide/admin.md
```diff
@@ -19,6 +19,79 @@ The Admin Panel is the Django Admin interface for the Aqueduct Gateway. It provi
 
 ![Admin Panel](../assets/user_guide/admin_panel.png)
 
+## Configuration Snippets
+
+Runtable OIDC/auth configuration (org extraction, OAuth team-mapping, and user
```
natkam: "Runtable" looks like a typo to me... Shouldn't it be "Runtime", perhaps?

docs/user-guide/admin.md
```diff
@@ -19,6 +19,79 @@ The Admin Panel is the Django Admin interface for the Aqueduct Gateway. It provi
...
+
+| Method | Signature | Failure behavior |
+|--------|-----------|------------------|
+| `org_name` | `(claims) -> str \| None` | fail closed (`None` ⇒ no login) |
+| `team_names` | `(claims) -> list[str]` | fail open (`[]`) |
+| `display_team_names` | `(team_names) -> list[(str, str)]` | identity mapping (names unchanged) |
+| `user_group` | `(claims) -> "admin" \| "org-admin" \| "user"` | fail open (`"user"`) |
+
+At most one `config` snippet may be active. With no active snippet, the default
+base class (today's out-of-the-box behavior: org `"default"`, no teams,
```
natkam: Which day does "today" refer to? I don't think this time reference makes sense here.

docs/user-guide/admin.md
```diff
@@ -19,6 +19,79 @@ The Admin Panel is the Django Admin interface for the Aqueduct Gateway. It provi
...
+`user` role) is used; in particular `display_team_names` defaults to an
+identity mapping that keeps each team's name unchanged. Saving a snippet that
+fails to validate (syntax errors,
+wrong method signatures, or a class that does not subclass `ConfigSnippet`) is
+rejected by the admin. Standard-library imports are permitted. The active
```
natkam: Is it rejected by the admin as a person, or by the admin form? This sentence sounds a bit weird.

How about something along the lines of:
```
A snippet must be valid Python code, with correct method signatures and a class
inheriting from `ConfigSnippet`.
```
This seems to be easier to read to me, but feel free to come up with some other variant.

docs/user-guide/admin.md
```diff
@@ -19,6 +19,79 @@ The Admin Panel is the Django Admin interface for the Aqueduct Gateway. It provi
...
+`user_group()` result (or if there is no snippet), so they can manage admin data
+without manual edits. Leave `ADMIN_GROUP` empty (default) to disable this and
+rely solely on the snippet's `user_group`.
+
+> **Security:** a snippet is arbitrary Python executed on the server. Editing is
+> **superuser-only** and there is **no sandboxing**. A test console is available
+> on the Snippets changelist to run a snippet against a sample input before
+> saving it; it executes real server code and is superuser-only.
+
+`ADMIN_SUPER_USER` was a bootstrap-only escape hatch (removed): admin
```
natkam: If it was removed, we don't need it in the docs, do we?

docs/user-guide/admin.md
```diff
@@ -42,52 +119,11 @@ OAuth team management automatically syncs user team memberships based on OAuth g
...
-2. OAuth groups are extracted from claims
-3. For each group, `OAUTH_TEAM_NAMES_FUNCTION` is called with the group name and full groups list
-4. Groups that return a tuple create teams; groups that return `None` are skipped
-5. Teams are created (if `ENABLE_OAUTH_GROUP_CREATION=True`) or reused
-6. User is added to teams via `TeamMembership`
-7. User is removed from teams no longer in their OAuth groups (only if `ENABLE_OAUTH_GROUP_REMOVAL=True`)
+The team-name mapping logic is provided by the active **config snippet's**
+`team_names` / `display_team_names` methods (see
+[Configuration Snippets](#configuration-snippets)); it is no longer set in
+`settings.py`.
```
natkam: This looks like it should be mentioned in the changelog (if we had one) and in the release notes, but I don't think we need references to the legacy behaviour in the docs.

aqueduct/aqueduct/settings.py
natkam: Looks like `ADMIN_SUPERUSER_EMAILS` is not documented anywhere though!

aqueduct/management/templates/admin/management/snippet/test_console.html
```diff
@@ -0,0 +1,108 @@
+{% extends "admin/base_site.html" %}
+{% load i18n static %}
+
+{% block extrastyle %}
+  {{ block.super }}
+  <link rel="stylesheet" href="{% static 'css/admin_custom.css' %}">
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/codemirror@5.65.16/lib/codemirror.min.css">
```
natkam: Not sure if this version is set here on purpose, but version 6 of CodeMirror is available and seems to be advertised by the creators of the package over the old version :) See e.g. the old repo's readme:
https://github.com/codemirror/codemirror5/

aqueduct/management/tests/test_snippet_admin.py
```diff
@@ -0,0 +1,276 @@
...
+            data={"name": "c2", "type": "config", "active": False, "code": VALID_CODE}
+        )
+        self.assertTrue(form.is_valid(), form.errors)
+        form.save()
+        self.assertEqual(Snippet.objects.filter(type=SnippetType.CONFIG, active=True).count(), 1)
+
+    def test_promoting_config_demotes_all_previous_active(self):
+        seed_active_config(SNIPPET_ORG_CUSTOM)
+        Snippet.objects.create(name="old1", type=SnippetType.CONFIG, active=True, code=VALID_CODE)
+        Snippet.objects.create(name="old2", type=SnippetType.CONFIG, active=True, code=VALID_CODE)
```
natkam, line 117: But doesn't the creation of the second active config automatically make the first one inactive? That's what the `test_orm_create_active_config_demotes_previous` below is testing. The name of the test and this setup here are somewhat misleading, because they suggest that you can actually have multiple active configs that would be disabled by saving the form.

And, in general, this test seems to duplicate what we're already testing in `test_second_active_config_demotes_previous`.

(I guess you could achieve what this test purports to be testing by using `bulk_create`, which doesn't call the model's `save()`, though!)

aqueduct/management/admin.py
```diff
@@ -881,3 +888,157 @@ def vector_store_link(self, obj) -> str:
...
+            "security_warning": SNIPPET_CONSOLE_SECURITY_WARNING,
+            "code": code,
+            "payload": payload,
+            "results": results,
+        }
+        request.current_app = self.admin_site.name
+        return TemplateResponse(request, "admin/management/snippet/test_console.html", context)
+
+    @staticmethod
+    def _run_console(snippet, payload: Any) -> list[str]:
```
natkam: Missing type annotation for `snippet` :)

aqueduct/management/admin.py
```diff
@@ -881,3 +888,157 @@ def vector_store_link(self, obj) -> str:
...
+            except ValidationError as e:
+                self.message_user(request, f"Snippet failed to compile: {e}", messages.ERROR)
+            else:
+                parsed = {}
+                if payload.strip():
+                    try:
+                        parsed = json.loads(payload)
+                    except json.JSONDecodeError as e:
+                        self.message_user(request, f"Invalid JSON test input: {e}", messages.ERROR)
+                if parsed is not None:
```
natkam: Hmm, shouldn't it be `if parsed != {}`, by any chance? Or, actually, shouldn't `_run_console` only run when there was no `JSONDecodeError`?

aqueduct/management/auth.py
```diff
@@ -93,24 +76,37 @@ def _get_teams(self, team_names: list[str]) -> list[tuple[str, str]]:
...
+        admin_group = getattr(settings, "ADMIN_GROUP", "")
+        if admin_group and admin_group in team_names:
+            group = UserGroup.ADMIN
 
-        Note: admin determination uses the raw team/group name extracted from
-        the claims (same as team sync), not the display team name.
-        """
-        is_admin = bool(getattr(settings, "ADMIN_GROUP", None) in team_names)
+        admin_superuser_emails = getattr(settings, "ADMIN_SUPERUSER_EMAILS", [])
+        is_superuser_email = bool(user.email and user.email.lower() in admin_superuser_emails)
```
natkam: Unnecessary cast to bool:

```suggestion
        is_superuser_email = user.email and user.email.lower() in admin_superuser_emails
```

aqueduct/management/tests/test_snippet_admin.py
```diff
@@ -0,0 +1,276 @@
...
+        active = Snippet.objects.filter(type=SnippetType.CONFIG, active=True)
+        self.assertEqual(active.count(), 1)
+        self.assertEqual(active.get().name, "c2")
+
+    def test_two_active_plugins_same_name_allowed(self):
+        Snippet.objects.create(name="plug", type=SnippetType.PLUGIN, active=True, code=VALID_CODE)
+        form = SnippetAdminForm(
+            data={"name": "plug", "type": "plugin", "active": True, "code": VALID_CODE}
+        )
+        self.assertTrue(form.is_valid(), form.errors)
```
natkam, line 134: You could maybe add an assertion about the objects created in the database: that there are two plugins, and both are active.

aqueduct/management/auth.py
```diff
@@ -93,24 +76,37 @@ def _get_teams(self, team_names: list[str]) -> list[tuple[str, str]]:
 
         return team_mappings
 
-    def _update_user_role(self, user: User, profile: UserProfile, team_names: list[str]):
-        """
-        Sync the user's role from the raw OAuth team (group) names.
+    def _update_user_role(
+        self, user: User, profile: UserProfile, team_names: list[str], claims
```
natkam: Missing type annotation for `claims` :)

aqueduct/management/tests/test_oauth_user_role.py
```diff
@@ -8,84 +8,224 @@
...
+        self.backend._update_user_role(user, profile, ["admins"], {"groups": ["admins"]})
+
+        profile.refresh_from_db()
+        user.refresh_from_db()
+        self.assertEqual(profile.group, "admin")
+        self.assertTrue(user.is_staff)
+        self.assertTrue(user.is_superuser)
+
+    @override_settings(ADMIN_SUPERUSER_EMAILS=["you@example.com"])
+    def test_dev_auto_admin_email_forced_superuser_regardless_of_group(self):
```
natkam: What does `dev_auto` refer to here and in the two next tests' names?

natkam (changes requested):
Some nit-picks and minor improvement suggestions only. Overall this really looks great! 🚀

aqueduct/aqueduct/settings.py
JaeYeonLee0621: Thanks!! I added the explanation in admin.md documents, this part **Auto-admin via `ADMIN_SUPERUSER_EMAILS`**

aqueduct/management/templates/admin/management/snippet/test_console.html
```diff
@@ -0,0 +1,144 @@
...
+        </ul>
+      </fieldset>
+    {% endif %}
+  </div>
+</div>
+
+<!-- CodeMirror 6 (ES modules) via import map. Versions pinned to the exact
+     closure resolved by npm install codemirror@6.0.2 lang-python lang-json. -->
+<script type="importmap">
+{
```
JaeYeonLee0621, line 90: I upgraded codemirror version 5 to 6. I don't know about the frontend that well so work with AI :)

aqueduct/management/admin.py
```diff
@@ -881,3 +888,150 @@ def vector_store_link(self, obj) -> str:
...
+                snippet = compile_snippet(code)
+            except ValidationError as e:
+                self.message_user(request, f"Snippet failed to compile: {e}", messages.ERROR)
+            else:
+                if payload.strip():
+                    try:
+                        parsed = json.loads(payload)
+                    except json.JSONDecodeError as e:
+                        self.message_user(request, f"Invalid JSON test input: {e}", messages.ERROR)
+                    else:
```
JaeYeonLee0621, line 987: I change this if condition to `else` to make sure that no `JsonDecodeError` parsed data go to the `_run_console`.

docs/user-guide/admin.md
```diff
@@ -15,25 +15,109 @@ nav_order: 8
 
 ---
```
JaeYeonLee0621, line 17: Here, I changed some typo and remove some unnecessary or too much detail explanation. 
Also edit this file alignment little bit.

natkam (approved):
Awesome! Feel free to merge it whenever you have time :)

JaeYeonLee0621:
Thanks for the review! :) I'll merge it once Matthias checks it

PR #185, opened 2026-08-26
152: More efficient http requests processing

### LINKED ISSUE #152: More efficient request post-processing
Currently all endpoints return a Django `JsonResponse` object in which the JSON data is already serialized. Modifying this data requires parsing the JSON again.

There are many paths forward. We could create a sub-class where the data is not serialized immediately. We might return the raw data and introduce a new decorator that finally wraps the response in a `JsonResponse` etc. etc.

natkam:
**Main goal**

The gateway views no longer return Django `StreamingHttpResponse`/`JsonResponse` objects directly. Instead they return lightweight "raw" response objects, containing a Pydantic model or a dict as data, and a newly added `HttpResponseMiddleware` middleware converts those raw objects into real HTTP responses. (Exceptions: the two cases when there is no json in the response content: the speech endpoint, which streams bytes in a `StreamingHttpResponse`, and transcriptions with text format, which returns an `HttpResponse` with text content type.)

For streaming responses, in order to avoid repeatedly iterating over the content to apply any post-processing and then creating a new async generator, the post-processing transformations are gathered in a list (`RawStreamingResponse.transforms`) and then applied to individual chunks in the middleware. Example: `normalize_reasoning_fields()` decorator.

Note: MCP streaming responses have to yield a slightly different format of data then other streaming responses, and they do not have a `request_log` object attached, so the middleware handles them separately from the others.

The idea here is that any post-processing of the response can be now executed on a Python object, without repeatedly converting the response's content to JSON and back. The conversion happens only once, at the very end of the response handling - in the middleware.

**Other unrelated changes**
- MCP tests can now run without internet connection, providing the server package is already installed (`--prefer-offline` flag).
- Removed an unused helper method for tests (mock router).

### DIFF
--- a/aqueduct/gateway/middleware.py
+++ b/aqueduct/gateway/middleware.py
@@ -6,7 +6,7 @@ from functools import reduce
 from typing import Any, TypeVar
 
 from django.core.handlers.asgi import ASGIRequest
-from django.http import JsonResponse, StreamingHttpResponse
+from django.http import HttpResponse, JsonResponse, StreamingHttpResponse
 from litellm.types.utils import ModelResponseStream
 
 from gateway.views.utils import RawJsonResponse, RawStreamingResponse, get_token_usage
@@ -15,6 +15,7 @@ from management.models import Usage
 log = logging.getLogger("aqueduct")
 
 
+ViewResult = HttpResponse | StreamingHttpResponse | RawJsonResponse | RawStreamingResponse
 T = TypeVar("T", bound=ModelResponseStream | dict[str, Any])
 
 
@@ -76,16 +77,15 @@ class HttpResponseMiddleware:
     Transform raw responses from the gateway views into valid HTTPResponses.
 
     This middleware should be applied after all data post-processing
-    middleware.
+    middleware. Note: expects ``get_response`` to be synchronous (views are
+    already wrapped by upstream middleware).
     """
 
-    def __init__(
-        self, get_response: Callable[[ASGIRequest], JsonResponse | StreamingHttpResponse]
-    ) -> None:
+    def __init__(self, get_response: Callable[[ASGIRequest], ViewResult]) -> None:
         self.get_response = get_response
 
-    def __call__(self, request: ASGIRequest) -> JsonResponse | StreamingHttpResponse:
-        """Transform a raw response from a gateway view into a valid HTTPResponse.
+    def __call__(self, request: ASGIRequest) -> JsonResponse | StreamingHttpResponse | HttpResponse:
+        """Transform a raw response from a gateway view into a valid HttpResponse.
 
         If the response is not an instance of ``RawJsonResponse``
         or ``RawStreamingResponse``, it is returned unchanged.
--- a/aqueduct/gateway/middleware.py
+++ b/aqueduct/gateway/middleware.py
@@ -1,0 +1,87 @@
+from collections.abc import Callable
+from typing import Any, reveal_type
+
+from asgiref.sync import iscoroutinefunction
+from django.core.handlers.asgi import ASGIRequest
+from django.http import JsonResponse, StreamingHttpResponse
+from django.utils.decorators import sync_only_middleware
+
+from gateway.views.utils import RawJsonResponse, RawStreamingResponse, _openai_stream
+
+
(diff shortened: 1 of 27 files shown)


natkam:
@JaeYeonLee0621 Since I'm away until 23.09, please feel free to do whatever you need with this PR, especially if it blocks you from working on other tasks! Otherwise, if you have any comments, I'll apply them when I'm back :)

aqueduct/gateway/views/image_generation.py
```diff
@@ -73,7 +72,7 @@ async def image_generation(
             "Unexpected argument in request body", pydantic_model.get("model"), llm_provider=None
         ) from err
 
-    data = resp.model_dump(exclude_unset=True)
-    request_log.token_usage = _get_token_usage(data)
+    # TODO # data = resp.model_dump(exclude_unset=True)
```
JaeYeonLee0621, line 75: Do we still need this comment? It looks like it might be left over from before the change.

aqueduct/gateway/views/vector_stores.py
```diff
@@ -153,7 +153,7 @@ async def vector_stores(
     # Return upstream response directly (ID already matches)
     response_data = remote_vs.model_dump(mode="json")
```
JaeYeonLee0621, line 154: Could we pass `remote_vs` here instead? If you agree, the logic below would need the same change.

aqueduct/gateway/views/utils.py
```diff
@@ -166,31 +197,19 @@ def __init__(self, streaming_content: AsyncGenerator[str, None], model: str, ema
...
-    async def __anext__(self) -> str:
-        chunk: str = await self.streaming_content.__anext__()
-        if not self._registered and chunk:
-            response_id = self.extract_response_id_from_chunk(chunk)
+    async def __anext__(self) -> ResponseStreamEvent:
+        chunk: ResponseStreamEvent = await self.streaming_content.__anext__()
+        if (
+            not self._registered
+            and isinstance(chunk, ResponseCreatedEvent)
+            and chunk.type == "response.created"
```
JaeYeonLee0621, line 205: Since `ResponseCreatedEvent` already narrows `chunk.type` to `response.created`, is this condition still needed?

```python
type: Literal["response.created"]
```

aqueduct/gateway/middleware.py
```diff
@@ -0,0 +1,130 @@
...
+        if isinstance(response, RawJsonResponse):
+            # Merge headers from response.headers (may have been modified after init)
+            kwargs = response.kwargs.copy()
+            kwargs["headers"].update(response.headers)
+
+            if isinstance(response.content, BaseModel):
+                response.content = response.content.model_dump(
+                    exclude_none=True, exclude_unset=True, mode="json"
+                )
+            else:
```
JaeYeonLee0621, line 105: Could we change `model_dump()` to `model_dump(mode='json')` here for consistency? Same for the logic below, if that works for you.

- `mode="python"` (default) : keeps native Python types
- `mode="json"` : converts to JSON-compatible types

aqueduct/gateway/views/utils.py
```diff
@@ -1,89 +1,120 @@
...
 from management.models import Request, Usage
 
 log = logging.getLogger("aqueduct")
 
+T = TypeVar("T", bound=ModelResponseStream | JSONRPCMessage)
 
-def _get_token_usage(content: bytes | dict[str, Any]) -> Usage:
-    """Retrieves token usage information from the response content.
+
+class RawJsonResponse:
```
meffmadd, line 34: `RawJsonResponse` is intended to replace `JsonResponse` so why not sub-class it? 

I had a quick look at the implementation and it seems like Django probably uses the `content` property to determine what data to send. The only problem with the current `JsonResponse` is that it immediately `json.dumps` the content. We could just rewrite the class to dump the JSON content dynamically as far as I understand it. I assume Django just accepts a sub-class as response type.

This would make the middleware obsolete and would be cleaner I think.

meffmadd (changes requested):
Can you check the feasibility of the sub-classing approach? Maybe I missed something but subclassing seams simpler to me...

PR #187, opened 2026-09-01
169: Remove old Bitnami chart references and values

### LINKED ISSUE #169: Remove old Bitnami chart references and values
We still reference the old Bitnami charts in the Helm chart. We removed it on our deployments so they now become dead weight.

These can be removed after a grace period for other deployments. Replacement of Bitnami charts happened 16.7.2026.

JaeYeonLee0621:
Bitnami's postgresql and redis charts are outdated and no longer maintained, so       
 they've been removed.                                                                 
                                                                                       
 - PostgreSQL → now runs via CloudNativePG (CNPG) to run PostgreSQL in Kubernetes.     
 - Redis → replaced by Valkey (a drop-in Redis-compatible fork).                       
                                                                                       
 This removes all Bitnami and outdated code, updates the related comments, and deletes 
 the vendored .tgz files.


charts/aqueduct/values.yaml
```diff
@@ -185,51 +160,34 @@ cnpg:
...
-  # allow replication connections (wal_level >= replica, a pg_hba `replication`
-  # rule, and a user with REPLICATION privilege, e.g. the `postgres` superuser).
-  # Only takes effect on first creation (the Cluster must not yet exist).
+  # PostgreSQL instance via pg_basebackup, instead of running initdb. CNPG clones
+  # the source at creation time and then promotes the clone to a primary, so no
+  # separate dump/restore job is needed. The source must allow replication
+  # connections (wal_level >= replica, a pg_hba `replication` rule, and a user
+  # with REPLICATION privilege, e.g. the `postgres` superuser). Only takes effect
+  # on first creation (the Cluster must not yet exist).
   bootstrapFromExternal:
```
JaeYeonLee0621, line 169:
```yaml
bootstrapFromExternal:
    enabled: true
    connection:
      # The `postgres` superuser password lives in the same secret Bitnami
      # reads (global.postgresql.auth.secretKeys.adminPasswordKey).
      existingSecret: "aqueduct-secrets"
      secretKeys:
        password: "POSTGRES_PASSWORD"
```
Is `bootstrapFromExternal` necessary in `aqueduct/staging.yaml`? If not, can I delete it?

+) I'm guessing `bootstrapFromExternal` was only used the first time CNPG started to clone the old database and prevent data loss.

+) https://gitlab.tuwien.ac.at/data-science/chart-values/-/blame/ai-platform/aqueduct/staging.yaml?ref_type=heads#L117

charts/aqueduct/values.yaml
```diff
@@ -147,31 +142,11 @@ ingress:
   tls: false
   tlsSecretName: ""
 
-postgresql:
```
JaeYeonLee0621, line 150: Can I remove this block from `aqueduct/staging.yaml` as well?

```yaml
postgresql:
  enabled: false  # Step 3 of Bitnami -> CNPG migration: stop Bitnami, app auto-derives CNPG host/creds. PVC is NOT deleted.
  primary:
    extraEnvVars:
      - name: POSTGRESQL_MAX_CONNECTIONS
        value: "300"
    extendedConfiguration: |
      shared_buffers = 6000MB
    # Allow pg_basebackup replication connections from the CNPG pod (the
    # default pg_hba.conf has no `host replication` rule, and the `all`
    # database does not match the `replication` pseudo-database). Required for
    # the side-by-side CNPG bootstrap in step 1 of the Bitnami -> CNPG migration.
    pgHbaConfiguration: |-
      local   all             all                                     trust
      host    all             all             0.0.0.0/0               md5
      host    all             all             ::/0                     md5
      host    replication     postgres        0.0.0.0/0               md5
      host    replication     postgres        ::/0                     md5
    persistence:
      enabled: true
      size: 10Gi
    resources:
      limits:
        cpu: "2"
        memory: "8Gi"
      requests:
        cpu: "2"
        memory: "8Gi"
```

+) https://gitlab.tuwien.ac.at/data-science/chart-values/-/blame/ai-platform/aqueduct/staging.yaml?ref_type=heads#L76

natkam:
To me it looks good, but you could also ask Josef to have a look :)

charts/aqueduct/values.yaml
meffmadd: Yes this can be removed once it is merged.

charts/aqueduct/values.yaml
meffmadd: Yes we can disable it but leave the config for posterity as documentation.

meffmadd (commented):
Changes look reasonable but we should wait some time before merging this.

charts/aqueduct/values.yaml
JaeYeonLee0621: Can I open an issue so I don't forget it?

JaeYeonLee0621:
> Changes look reasonable but we should wait some time before merging this.

Just curious, what's blocking this PR from being merged?

meffmadd:
We should provide some deprecation period for other users as the charts would stop working if there is no CNPG operator installed on the cluster.

PR #188, opened 2026-09-11
129: Refactoring decorators

JaeYeonLee0621:
### 1. Decorators refactored

Split the large `gateway/views/decorators.py `into a new `gateway/decorators/` package with one module per concern:

- auth.py — token_authenticated, tos_accepted, check_limits
- body.py — parse_body, FileSizeError, ensure_usage, resolve_alias
- availability.py — check_model_availability
- log.py — log_request
- errors.py — catch_router_exceptions
- mcp.py — check_mcp_server_availability, mcp_transport_security, parse_jsonrpc_message
- responses.py — validate_response_id, check_tool_availability and helpers
- files.py — require_files_api_client, process_batch_file and its helpers
- chat_completions.py — process_file_content, normalize_reasoning_fields and their helpers
- types.py — shared type aliases
- init.py — re-exports the decorators as a single facade

The old `gateway/views/decorators.py` was deleted.

### 2. Review opinions applied

- `check_mcp_server_availability` moved into `mcp.py`.
- `get_relay_model_name` removed (it was unused).
- `process_file_content` and `normalize_reasoning_fields` grouped in `chat_completions.py.`
- `Utility functions` placed next to the decorators that use them, rather than in a separate utils module.


aqueduct/gateway/decorators/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+from gateway.decorators.files import process_batch_file, require_files_api_client
+from gateway.decorators.log import log_request
+from gateway.decorators.mcp import (
+    check_mcp_server_availability,
+    mcp_transport_security,
+    parse_jsonrpc_message,
+)
+from gateway.decorators.responses import check_tool_availability, validate_response_id
+
+__all__ = [
```
meffmadd: Should we use `__all__` here? I guess we just assume we want to export everything and we still have private functions with `_`. This is always more work with little gain.

aqueduct/gateway/decorators/auth.py
```diff
@@ -0,0 +1,235 @@
...
+                        "In order to use the API you have to agree to the terms of service!",
+                        status=403,
+                    )
+
+        return await view_func(request, *args, **kwargs)
+
+    return wrapper
+
+
+def check_limits(view_func: AsyncView) -> AsyncView:
```
meffmadd: I think `check_limits` should not belong to auth. As we already have plans to expand how we handle limits, this should get its own file I think.

aqueduct/gateway/decorators/chat_completions.py
```diff
@@ -0,0 +1,220 @@
...
+    if reasoning_content is not None and reasoning is None:
+        message["reasoning"] = reasoning_content
+        return True
+    if reasoning is not None and reasoning_content is None:
+        message["reasoning_content"] = reasoning
+        return True
+    return False
+
+
+def normalize_reasoning_fields(view_func: AsyncView) -> AsyncView:
```
meffmadd, line 182: Tangential but do we really only need to normalize reasoning fields for chat completions and not responses? I forgot but maybe the Responses was unified from the beginning anyways.

aqueduct/gateway/response_types.py
meffmadd, line 1: I do not really like the name of the file. Should we have a `types.py` file? Or leave it in `utils.py`? Basically can you check how many custom utility classes and functions we have/expect to have and split accordingly?

aqueduct/gateway/raw_response.py
```diff
@@ -0,0 +1,138 @@
...
+        422: "invalid_request_error",
+        429: "rate_limit_error",
+        500: "server_error",
+        503: "server_error",
+        504: "timeout_error",
+    }
+    return status_map.get(status, "invalid_request_error")
+
+
+def in_wildcard(value: str | None, allowed_values: list[str]) -> bool:
```
meffmadd: This is probably a util function right?

aqueduct/gateway/raw_response.py
```diff
@@ -0,0 +1,138 @@
...
+        # Check wildcard port patterns (e.g., "http://localhost:*")
+        for allowed in allowed_values:
+            if allowed.endswith(":*"):
+                base_origin = allowed[:-2]
+                if value.startswith(base_origin + ":"):
+                    return True
+    return valid
+
+
+def register_response_in_cache(response_id: str | None, model: str, email: str) -> None:
```
meffmadd: This and the following functions are specific to the Responses API and not related to Django HTTP responses at all so these should be moved.

meffmadd (changes requested):

aqueduct/gateway/decorators/__init__.py
```diff
@@ -0,0 +1,21 @@
+from gateway.decorators.auth import token_authenticated as token_authenticated
```
JaeYeonLee0621: mypy treats imports in a module as private by default, so re-exporting a name requires the import X as X form to mark it as public.

aqueduct/gateway/decorators/__init__.py
JaeYeonLee0621: Agreed, `__all__` isn't needed here, so I've removed it. External code can import the decorators using their full path.

aqueduct/gateway/decorators/auth.py
JaeYeonLee0621: I put the `check_limits` definition in a new `limits.py` file. I'm open to suggestions on the file name or placement :)

aqueduct/gateway/decorators/chat_completions.py
JaeYeonLee0621: I'm not sure I fully follow this comment. Could you explain it a bit more?

aqueduct/gateway/response_types.py
JaeYeonLee0621: I renamed the file to `response_type.py`, since putting all these classes in a `utils.py` file didn't feel right and `types.py` seemed too generic. Let me know if you'd prefer something else and I'll change it.

aqueduct/gateway/raw_response.py
JaeYeonLee0621: I moved this function to `utils.py`.

aqueduct/gateway/raw_response.py
JaeYeonLee0621: I moved the cache-related definitions into `response_cache.py`.

aqueduct/gateway/decorators/chat_completions.py
meffmadd: There were inconsistencies with the different Chat Completions API implementations where some used `reasoning` and others  `reasoning_content` so we just make sure both are in the response. I was wondering if the Responses API has the same issue.

aqueduct/gateway/raw_response.py
meffmadd: I think they should be in views/responses.py since it is so specific to the Responses endpoint. But I am not sure.

aqueduct/gateway/response_types.py
meffmadd: Yes ok that makes sense! Also note that this file heavily depends on `152-http-requests-processing` and I wanted to check if we could rewrite the classes so that we can get rid of the middleware introduced there. So we might have to merge this first otherwise we have to merge again here etc.

aqueduct/gateway/decorators/__init__.py
meffmadd: We also use it in the model refactor so you can also leave it... sorry that was probably a bit premature.

aqueduct/gateway/decorators/chat_completions.py
JaeYeonLee0621: I'm not familiar with this part, but from what AI said, the `Responses API` path doesn't go through LiteLLM's routing.  It calls `client.responses.create()` directly on an OpenAI-compatible SDK, and the result comes back as a canonical, strongly typed Response pydantic model.  So It said, It doesn't have the same issue as the Chat Completions API.

aqueduct/gateway/raw_response.py
JaeYeonLee0621: I tried making this change, but it runs into two circular imports.

Before

```
responses_cache.py (register/get/delete) → decorators/responses.py (validate_response_id)
responses_cache.py (register/get/delete) → views/responses.py (create_response) 
responses_cache.py (register/get/delete) → views/utils.py (ResponseRegistrationWrapper) 
```

After (functions moved to views/responses.py): 

```
decorators/responses.py (validate_response_id) <-> views/responses.py (get_response_from_cache)
views/utils.py(ResponseRegistrationWrapper) <-> views/responses.py (register_response_in_cache)
```

PR #189, opened 2026-09-14
Fix service account edit page 500 Error

JaeYeonLee0621:
## Problem

Service account edit page returns a 500 (`ImproperlyConfigured`). 

- Production

Screenshot from 2026-09-14 11-56-19

- Local

Screenshot from 2026-09-14 11-56-07

## Cause
`ServiceAccountUpdateView` inherits from `UpdateView` without declaring `form_class` or `template_name`, so every request to `/service-accounts/<id>/edit/` fails before rendering.     

## Fix
Set `form_class` and `template_name` on the update view, plus a regression test covering GET and POST. 

Screenshot from 2026-09-14 12-03-46


meffmadd (approved):
LGTM!

PR #191, opened 2026-09-15
129: Refactoring models

JaeYeonLee0621:
1. Split the `aqueduct/management/models.py` into a `management/models/` package

- models/__ init __.py — re-exports every model/helper
- models/mixins.py — LimitSet, LimitMixin, ModelExclusionMixin, MCPServerExclusionMixin 
- models/accounts.py — Org, Team, UserGroup, UserProfile, TeamMembership, ServiceAccount, Token 
- models/snippets.py — SnippetType, Snippet 
- models/usage.py — Usage, Request
- models/files.py — FileObject
- models/batches.py — default_request_counts, BatchStatus, Batch
- models/vector_stores.py — VectorStore/VectorStoreFile/VectorStoreFileBatch + status enums 

2. Deleted the old `models.py`
 
+) No DB changes → no migration needed.
+) Historical migrations still resolve via the package __init__.py


aqueduct/management/models/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+from management.models.vector_stores import (
+    VectorStore,
+    VectorStoreFile,
+    VectorStoreFileBatch,
+    VectorStoreFileBatchStatus,
+    VectorStoreFileStatus,
+    VectorStoreStatus,
+)
+
+__all__ = [
```
meffmadd: I generally dislike `__all__` because it is always a hassle to update for little gain but if you have a case for it it is fine.

aqueduct/management/models/user_profile.py
meffmadd, line 1: I feel like `accounts.py` is an unfitting name for this... How about `membership` or `identity`?

aqueduct/management/models/accounts.py
```diff
@@ -0,0 +1,475 @@
...
+                        }
+                    )
+
+            except Team.DoesNotExist:
+                # This case shouldn't happen if ForeignKey validation runs,
+                # but good to handle defensively.
+                raise ValidationError("Associated team does not exist.") from None
+
+
+class Token(models.Model):
```
meffmadd: Token should have its own file I think... this file is already very long again.

meffmadd (changes requested):
When I planned this I would have just created a single file for each `models.Model`. I think this is still the simplest solution. Some classes are long anyways and the naming is also straightforward then.

aqueduct/management/models/user_profile.py
JaeYeonLee0621: `identity` sounded like it refers to a single account. Since this file covers service accounts, organizations, profiles, and so on, I named it `membership.py`.

aqueduct/management/models/accounts.py
JaeYeonLee0621: I created a separate `tokens.py` file for this model.

aqueduct/management/models/__init__.py
JaeYeonLee0621: I kept `__all__` here because

1. Dropping these re-exports would force full paths everywhere and break historical migrations (e.g. `management/migrations/0005_file_purpose_update.py`).
2. The "as" re-export (from X import Y as Y) works , but `__all__` is the pattern already used (e.g. `gateway/views/__init__.py`).

meffmadd: Ok got it! Keep it as is then.

meffmadd (changes requested):
To make things as simple as possible can you just create a new file for each model with the file having the same name as the class? I think that does make sense here and we do not have to worry about naming.

JaeYeonLee0621:
I've applied your comments and moved each model into its own file, named after the class. :)

